PR #2, opened 2026-04-13
[Feat] Base Models + Schema + Seed Rakes

lukechanning:
## Summary

Implements the complete data model and schema layer for the water quality data explorer. This covers all 16 database tables, 15 ActiveRecord models with associations and validations, the `Filterable` concern powering the filter API, FactoryBot factories, model specs, and a state-scoped rake task for seeding development data from S3-sourced CSVs.

## Changes

- **Database schema** — 16 migrations in dependency order: PostGIS extension, `public_water_systems` (string PK `pwsid`), 8 domain tables hanging off it (`service_area_geometries`, `demographics`, `violations_summaries`, `environmental_justices`, `funding_summaries`, `watershed_hazards`, `boil_water_summaries`, `trend_data`), `tile_cache` (composite PK), `data_imports`, `place_system_crosswalks` (unique composite index on `geoid/pwsid`), and 3 cartographic reference tables (`states`, `counties`, `places`) with PostGIS geometry columns and GiST indexes

- **Models** — all 15 ActiveRecord model classes with correct primary key config (`self.primary_key = "pwsid"` on parent; explicit `foreign_key` + `primary_key` on all child `belongs_to`), `has_one`/`has_many :through` associations, and presence/format validations on `PublicWaterSystem`

- **`Filterable` concern** — `apply_filters(params)` class method on `PublicWaterSystem` supporting: categorical (AND between groups, OR within), boolean, direct range, state/geographic, join-based range filters across all 7 associated tables (violations, demographics, EJ, funding, hazard, trend, BWN), place/county geographic filters, and PostGIS bounding box intersection; uses `left_join_once` dedup pattern with a `Set` to prevent duplicate LEFT JOINs when multiple filters hit the same table; numeric coercion guards against non-numeric range params

- **Development seed task** — `db:seed:states[VT,RI,CT]` rake task loads 8 domain tables from filtered CSVs (VT/RI/CT rows only, ~820KB vs 45MB national); handles EJ data consolidation from 4 source files and watershed hazard aggregation by `pwsid`; helpers scoped to a `SeedImport` module to avoid polluting global `Object`

- **Seed CSVs** — 11 CSV files pre-filtered to VT, RI, and CT rows only committed under `db/seeds/csv/`

- **Test coverage** — 59 examples across 15 model specs covering associations, validations, all major `apply_filters` paths (categorical single/multi-select, boolean, range, join-based range, geographic via crosswalk), and `TileCache` composite PK uniqueness

## Notes for reviewers

`PlaceSystemCrosswalk` is intentionally not populated by the seed task — it's derived from spatial intersections between service area geometry polygons and cartographic places, which requires the ETL pipeline. The `place_geoid` filter therefore returns empty results in dev until ETL has run; this is expected behaviour, not a bug.

The `most_common_rate_tier` column in the demographics CSV has a typo at the source (`most_common_rate_tidy`) — the seed task maps it with a comment explaining this.


app/models/concerns/filterable.rb
CSheesley, line 1: So this is where the magic happens! I spent some time reading through this, I haven't used `Set` objects very often, but this is cool. I've also always dug this explicit pattern of building of a scope - love it!

Since this is the crux of the app, I wanted to make sure that we are confident on the testing here. In asking Claude about this, this is what it said:
```
Test coverage for Filterable#apply_filters

Current state: the .apply_filters block in public_water_system_spec.rb covers the happy path for a handful of filter types (categorical, boolean, direct range, one violation filter, one demographic filter, place geographic), but has meaningful gaps.

Missing coverage:

County geographic filter (county_geoid) — the PostGIS spatial subquery, arguably the riskiest code in this concern
Bounding box filter (bounds) — also spatial, also untested
Boil water, funding, EJ, watershed hazard, and trend filter groups — zero tests each
The duplicate-join guard (left_join_once) — no test verifying that activating two filter groups that share a table doesn't produce a broken query
Recommendations:

Move all apply_filters tests to a dedicated spec/models/concerns/filterable_spec.rb — the concern is large enough to warrant its own file, and it keeps public_water_system_spec.rb focused on validations/associations.

Add at least one test per filter group — enough to confirm the join is applied and the where clause fires correctly. We don't need to test every column in a group, just that the group works end-to-end.

Add one integration-style test that combines filters from multiple join-dependent groups (e.g. a violations filter + a demographics filter active simultaneously) to verify left_join_once keeps the query valid.

Add tests for both geographic filters — these touch PostGIS functions and are the most likely to break silently if the geometry data or spatial index changes.

The goal is confidence that any combination of filter params produces a valid, correct query — not exhaustive column-by-column coverage.
```

This response also points out another thing I was wondering about. While it appears that it is conventional to test concerns inside the model spec for the model that includes the concern - since this is a 1st class part of our app, and would have a pretty large set of associated tests...

What are your thoughts of moving the current specs for the concern into a new `spec/models/concerns/filterable_spec.rb` file - along with adding more specs to this file (based on the review and suggestion from Claude above) for full confidence and organization?

CSheesley (approved):
I think this looks great, and really appreciate you getting the foundation laid!!!

Aside from some of the ideas we chatted about over slack _(StandardRB, Annotations, and adding a glossary)_, I just left a comment and possible suggestion around the `Filterable` concern. Feel free to ignore it - not a blocker!

---
Oh!
What are your thoughts on some of the CSV document names, like: `cejst.csv`, `cvi.csv`, etc. ? Do those names match the actual names that EPIC uses when they upload files? If so, maybe we can make sure some of these things are listed in the Glossary? If we get to name these, maybe we can make some of the more explicitly named?
- Also not a blocker or major concern.

app/models/concerns/filterable.rb
lukechanning: Truly great callout @CSheesley ! Feedback should be addressed in these changes: 

https://github.com/bitfootco/water-data-tool/pull/2/commits/e224b80d50f03b23f5034abf66ce905c3281eb65

We can always expand on this in the future as well, but totally an amazing thing to flag.

PR #5, opened 2026-04-15
[Feat] PWS API — index, show, and export endpoints (M3)

CSheesley:
## Summary

Implements the `PublicWaterSystem` JSON API: index, show, and export endpoints. All ~50 filter params are wired through the existing `Filterable` concern. Server-side pagination via Pagy replaces the legacy DataTables client-side approach. CSV and gzip-compressed GeoJSON export are filter-respecting and unpaginated.

## Changes

- **API endpoints**
  - `GET /public_water_systems` — filtered, sorted, paginated index with a summary block (total systems, population served, open violations) that operates on the full filtered scope regardless of page
  - `GET /public_water_systems/:pwsid` — detail show with all 8 `has_one` associations; returns `null` for unpopulated associations rather than raising
  - `GET /public_water_systems/export` — CSV (default) or gzip-compressed GeoJSON; respects all active filters; bypasses pagination entirely

- **Serializers** — PORO classes (no gem); index serializer covers the ~15 table-view fields, detail serializer covers all associations

- **Exporter** — `PublicWaterSystemExporter` in `app/exporters/` handles both formats; ~70-column CSV matching the legacy export column layout, GeoJSON as a single-line `FeatureCollection`

- **`PublicWaterSystem.with_details` scope** — named scope for eager-loading all 8 associations, used by both show and export to avoid N+1s

- **`build_summary` query optimization** — combines `SUM` and `COUNT(*) FILTER (WHERE ...)` into a single PostgreSQL query via `Arel.sql` + `pick`; reuses `@pagy.count` for `systems_count`

- **Dev setup** — `dotenv-rails` added for `.env` loading in development/test; `DB_PORT` replaces `DATABASE_URL` to avoid Rails 12-factor config override; `seeds.rb` wired to call `db:seed:states` automatically; `annotaterb` reconfigured via generator output

## Notes for reviewers

- **`file_format` not `format`**: Rails reserves `params[:format]` for content negotiation — passing `format=geojson` would be treated as a route extension and attempt template rendering. Renamed to `file_format` throughout.

- **GeoJSON is a single line**: intentional. Pretty-printing adds ~30% overhead with no benefit for machine consumers. Matches legacy behavior.

- **CSV uses empty fields for nil, not `"NA"`**: the legacy export passed through raw `"NA"` strings from S3 source data. We cast at ETL time (`"NA"` → `NULL`) and emit blank cells, which downstream tooling (Excel, pandas, R) handles natively.

- **No HTML in CSV cells**: the legacy PHP export wrapped the EPA Facility Report URL in an `<a>` tag. We emit the plain URL.

- **`unscope(:order)` in `build_summary`**: the scope arrives with an `ORDER BY` from `apply_sort`. PostgreSQL rejects `ORDER BY` on an aggregate query without a `GROUP BY`. `unscope(:order)` strips it before the aggregation runs; filters are preserved.

- **Pagy v43**: the API changed significantly from older tutorials — `include Pagy::Method` in `ApplicationController`, `pagy(:offset, scope)` call signature, `Pagy::OPTIONS` for global config, `limit_key: "per_page"` to match our param convention.

- **Path 2 bulk downloads deferred to M13** — the Downloads section (pre-packaged S3 zips by state) is a separate milestone and not part of this PR.

## How to Manually Test _optional_

  Setup                                                                                                                                                                                      
  docker compose up -d                                                                                                                                                                       
  bin/dev                                                                                                                                                                                    
  mkdir -p tmp/test_exports                                                                                                                                                                  
                                                                    
  ---      
                                                                                                                                                                                  
  **Index - pagination**
  curl "http://localhost:3000/public_water_systems" | jq                                                                                                                                     
  curl "http://localhost:3000/public_water_systems?state=VT" | jq   
  curl "http://localhost:3000/public_water_systems?gw_sw_code=Groundwater&state=VT" | jq
  curl "http://localhost:3000/public_water_systems?page=2&per_page=10" | jq                                                                                                                  
  curl "http://localhost:3000/public_water_systems?sort_by=pws_name&sort_dir=asc&per_page=5" | jq
                                                                                                                                                                                             
  ---       
                                                        
  **Show - paginsation**                                                                                                                                                                                   
  curl "http://localhost:3000/public_water_systems/VT0020001" | jq  
  curl "http://localhost:3000/public_water_systems/DOESNOTEXIST" | jq  # → 404
                                                                                                                                                                                             
  ---

  **Export — CSV**                                                                                                                                                                               
  curl "http://localhost:3000/public_water_systems/export" -o tmp/test_exports/export.csv
  curl "http://localhost:3000/public_water_systems/export?state=VT" -o tmp/test_exports/vt_export.csv
  head -2 tmp/test_exports/vt_export.csv  # header row + first data row                                                                                                                      
                                                                                                                                                                                             
  ---                  
                                                                                                                                                                      
  **Export — GeoJSON**                                                                                                                                                                           
  _--compressed decompresses the gzip transport encoding, same as browser behavior_
                                                                                                          
  curl --compressed "http://localhost:3000/public_water_systems/export?file_format=geojson" -o tmp/test_exports/export.geojson
  curl --compressed "http://localhost:3000/public_water_systems/export?file_format=geojson&state=VT" -o tmp/test_exports/vt_export.geojson                                                   
  head -c 500 tmp/test_exports/export.geojson  # confirm FeatureCollection structure (file is single-line by design)


app/controllers/public_water_systems_controller.rb
```diff
@@ -0,0 +1,86 @@
...
+    total_pop, open_viol_count = scope.unscope(:order).pick(
+      Arel.sql("SUM(population_served_count)"),
+      Arel.sql("COUNT(*) FILTER (WHERE open_health_viol = 'Yes')")
+    )
+    {
+      systems_count: total_count,
+      total_population_served: total_pop,
+      systems_with_open_violations: open_viol_count
+    }
+  end
```
CSheesley, line 68: `.pick` runs a single aggregate query and returns the values directly — no model instantiation, no array unwrapping. Appropriate here because SUM and COUNT with no GROUP BY always collapse to exactly one row.

app/exporters/public_water_system_exporter.rb
CSheesley, line 1: Things I Learned: That Ruby uses the 'Exporter' pattern for things like this. I've always confusedly used a Service or a Helper - but this cool.

app/exporters/public_water_system_exporter.rb
```diff
@@ -0,0 +1,196 @@
...
+
+  # IMPORTANT: order and count must stay in sync with CSV_HEADERS above.
+  def csv_row(pws)
+    vs = pws.violations_summary
+    bws = pws.boil_water_summary
+    d = pws.demographic
+    td = pws.trend_datum
+    ej = pws.environmental_justice
+    fs = pws.funding_summary
+    wh = pws.watershed_hazard
```
CSheesley, line 86: I am not typically a fan of variable names like this - but it seems to make sense in many spots of our app, especially when the variable letters are a 1:1 match with the attribute name. Also keeps code clean that calls these variables.

config/database.yml
```diff
@@ -1,11 +1,11 @@
 default: &default
   adapter: postgis
   encoding: unicode
-  max_connections: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   host: localhost
-  port: 5432
-  username: postgres
-  password: postgres
+  port: <%= ENV.fetch("DB_PORT", 5432) %>
```
CSheesley, line 6: For some developers who have a Postgresql instance running in the background outside of docker for other projects... this allows somebody to define that port that docker will use. If we do not have that ENV var listed, we simply default to the default port 5432.

config/database.yml
```diff
@@ -16,6 +16,9 @@ test:
   database: water_data_tool_test
 
 production:
+  # TODO: Align production env var names with the DB_* convention used above
+  # (DATABASE_USERNAME → DB_USERNAME, WATER_DATA_TOOL_DATABASE_PASSWORD → DB_PASSWORD,
+  # DATABASE_HOST → DB_HOST). Requires updating Kamal config and server env vars in sync.
```
CSheesley, line 21: Do you have any thoughts on this? Not a big deal either way - wanted to add it as a TODO in case we want to come back to any spots.

lib/tasks/annotate_rb.rake
CSheesley, line 1: The previous`lib/tasks/annotate.rake` _(which may have been hand rolled?)_ was calling `do_annotations` with a plain Hash.

When trying to run a fresh migration after dropping and recreating the DBs, I would error out.

`annotaterb 4.x` — expects  `AnnotateRb::Options`. 

Fixed by running `bin/rails g annotate_rb:install` and letting the generator produce the boilerplate config for the version of the gem.

... @lukechanning  - I am curious if you would see an error message if you dropped a db and re-ran things. You may not even need to drop anything - and can just try `bin/rails db:migrate` on main... but I would assume you would have ran into the same error that I did at some point?

At any rate, running the install command and the two files that get generated _(what this pr has)_ seems to be the conventional path for using the `annotate_rb` gem.

spec/models/public_water_system_spec.rb
```diff
@@ -54,6 +54,22 @@
...
+      result = PublicWaterSystem.with_details.find_by(pwsid: pws.pwsid)
+
+      expect(result.association(:demographic)).to be_loaded
+      expect(result.association(:violations_summary)).to be_loaded
+      expect(result.association(:environmental_justice)).to be_loaded
+      expect(result.association(:funding_summary)).to be_loaded
+      expect(result.association(:watershed_hazard)).to be_loaded
+      expect(result.association(:boil_water_summary)).to be_loaded
+      expect(result.association(:trend_datum)).to be_loaded
+      expect(result.association(:service_area_geometry)).to be_loaded
```
CSheesley, line 69: TIL - the `.to be_loaded` assertion.

.annotaterb.yml
CSheesley, line 1: Default config for annotating with id values, followed by other values, followed by timestamps

Gemfile
```diff
@@ -10,6 +10,8 @@ gem "pg", "~> 1.1"
 gem "csv"
 # PostGIS adapter for ActiveRecord [https://github.com/rgeo/activerecord-postgis-adapter]
 gem "activerecord-postgis-adapter"
+# Pagination [https://github.com/ddnexus/pagy]
+gem "pagy", "~> 43.5"
```
CSheesley, line 14: `pagy` appears to be the common and conventional pagination gem used by RoR these days 🤷

Gemfile
```diff
@@ -45,6 +47,9 @@ gem "thruster", require: false
 gem "image_processing", "~> 1.2"
 
 group :development, :test do
+  # Load environment variables from .env [https://github.com/bkeepers/dotenv]
+  gem "dotenv-rails"
+
```
CSheesley, line 52: without this, ENV vars in the `.env` file were not being loaded when booting up the app.

README.md
```diff
@@ -58,21 +63,15 @@ Built for [EPIC](https://www.policyinnovation.org/) (Environmental Policy Innova
 
    This starts a PostGIS-enabled PostgreSQL container. The database runs on `localhost:5432`.
 
-5. **Set up the database**
+5. **Set up the databases**
 
    ```bash
    bin/rails db:setup
```
CSheesley, line 69: setup runs db: create, migrate, and seed - the seed file now kicks off the state seeding task _(and whatever other seeding we may want to add)_

docs/API.md
```diff
@@ -341,3 +340,79 @@ All errors return a JSON body:
 | 400 | Invalid filter parameters |
 | 404 | System not found (for `:pwsid` endpoints) |
 | 500 | Internal server error |
+
+---
+
+## Manual Testing
```
CSheesley, line 346: I wasn't sure of the best place to capture info like this. Should it even be captured and documented at all?

CSheesley (commented):
Ok - here it is - sorry it took a while 😄 

This PR matches the behavior of the currently deployed legacy app:
```
## Original Feature Behavior

**Path 1 — Table view export (filter-respecting)**
- User applies filters in the table view, clicks Export
- Radio buttons select `.csv` (default) or `.geojson`
- GeoJSON option is disabled when result count > 5,000 (legacy JS: `scripts.js:2610`)
- GeoJSON export is a standalone `.geojson` file — not a zip
- Gzip is HTTP transport only (`Content-Encoding: gzip`); browser decompresses before saving
- CSV is a flat ~70-column file across all associations
```

app/exporters/public_water_system_exporter.rb
lukechanning: Woah! I genuinely had no idea, that's so cool! 🤯

config/database.yml
lukechanning: Mm, a great catch. I'm so Docker-oriented, I don't often think enough about that. Great stuff!

config/database.yml
lukechanning: I'm actually all for this. I think env var inconsistency stacks up over time and just makes things "institutional knowledge" — it should be straightforward, I'm here for this

Gemfile
lukechanning: Sweet as, I wasn't familiar but a quick scan makes it seem like an obvious, quick win. PORO for pagination? Sure, hard to argue with that!

docs/API.md
lukechanning: Yup, I think any and all documentation is super duper welcome for this project. The fewer questions folks need to ask to get up and going, the better I reckon? Might eventually turn some of this into a public-facing "for the devs" page or something?

lukechanning (approved):
I'm gonna go ahead and merge this to keep crunching on my end, but absolutely stellar work @CSheesley ! As always, so thorough and well-constructed. Thanks for the effort!

PR #6, opened 2026-04-15
[Feat] Filter Bar, Detail Panel / Report, Tile Cache Invalidation (M11, M8, M6)

lukechanning:
## Summary

Implements milestones M11, M8, and M6 from the roadmap: tile cache invalidation after ETL imports, PWS detail popups with a printable report overlay, and filter bar interactivity (reset, URL sync, place autocomplete). Also fixes mobile layout so the app is usable on small screens.

### M11: Tile cache invalidation & warming
- Extract tile generation logic from `TilesController` into a standalone `TileGenerator` service module, slimming the controller to a thin HTTP wrapper
- Bust the tile cache after any successful ETL import (not just geometry imports) since MVT tiles embed non-geometry attributes like population and violation status
- Enqueue a `TileCacheWarmJob` that fans out per-zoom-level jobs (z0–z6) to pre-generate common tiles after cache bust, with per-layer fault isolation

### M8: PWS detail popup & printable report
- Add a Mapbox click popup on PWS polygons showing utility name, ID, state, type, connections, and population — with a "View Full Report" link
- Add `ReportsController` and report view that loads into a full-screen overlay via Turbo Frame, with print and close buttons
- Add HTML format to `PublicWaterSystemsController#show` with 8 data section partials (overview, demographics, EJ, violations, funding, watershed hazards, boil water, trends) shared between the detail page and the report view
- All tile property values in popups are HTML-escaped to prevent stored XSS

### M6: Filter bar completion
- Wire reset buttons on all filter menus — restores checkboxes, radios, selects, and population size toggles to defaults using `data-default` attributes
- Add URL sync: filter state is serialized to query params on apply (via `replaceState`) and restored from URL on page load for shareable filtered views
- Add place autocomplete search endpoint (`GET /places/search`) with prefix ILIKE matching, 10-result limit, LIKE special character escaping, and 1-hour cache headers
- Add `PlaceAutocompleteController` Stimulus controller with debounced fetch, DOM-constructed dropdown (XSS-safe), and mousedown-based blur handling
- Wire "Deselect all" toggle for ownership checkboxes and click-to-toggle for population size boxes
- Move `data-controller="filter"` to `#container-map` so Stimulus actions work in both the tab bar and the dropdown menu partials
- Add geocoder `result` event handler with context-aware zoom levels (state→z5, county→z7, city→z8) — the geocoder had `flyTo: false` with no handler wired

### Mobile layout
- Hide desktop filter tab bar, intro tooltip, zoom region buttons, and map/table toggle on screens ≤768px
- Size the map to fill the viewport between the fixed mobile header and footer
- Wire the hamburger menu toggle to show/hide the mobile menu overlay

## Notes for reviewers

- **Slider histograms deferred.** The plan called for replacing area/density `<select>` dropdowns with dual-handle sliders + histogram overlays. The dropdowns work and the sliders are a UX enhancement — deferred to avoid adding a noUiSlider dependency in this already-large branch.
- **Report feature is net-new.** The legacy app scaffolded a report overlay but never completed it (all JS was commented out, no data population). Our implementation is a new UX: click popup → "View Full Report" → Turbo Frame overlay.
- **Tile warm job scale.** z0–z6 is ~5,400 tile coordinates × 5 layers = ~27k SQL queries. Each zoom runs as a separate SolidQueue job. May need tuning with real data volumes.
- **Filter → map integration is M10.** Filters affect the DataTable (table view) immediately. Map tile filtering is a separate milestone not included in this branch.


CSheesley (approved):
I think I generally understand what is going on here - but maybe we chat about Caching a little to make sure I solidly understand how we are handling that topic in the app.

Do you think it is worth documenting 'important pieces' like this somewhere? In the `ETL.md` file, it seems like we have documentation around how that 'important piece' of the app works.  I did notice that it has a mention of busting the tile cache - maybe we can just add some info there explaining 'conceptually' our cache busting strategy, unles a full CACHE.md type file is warranted?

PR #8, opened 2026-04-15
[Feat] Add Stats Bar to Homepage

CSheesley:
## Summary

Implements the M7 **stats bar**: a Turbo Frame overlay showing aggregate stats (`systems count`, `customers`served, `area median income`, `open health violations`) that updates live as filters change. Also fixes three pre-existing filter bugs discovered during testing where the JS was sending values that didn't match what the database stores.

## Changes

**Stats bar (M7)**
- New `GET /public_water_systems/stats` endpoint renders a Turbo Frame HTML partial with 4 aggregate stats
- `build_summary` extended to include average median household income and systems count via a single PostgreSQL aggregate query (`SUM`, `COUNT(*) FILTER`, `AVG`, `COUNT DISTINCT` in one `.pick` call)
- `table_controller.js` wires up the stats frame: sets `frame.src` on `filters:changed`, deduplicates requests with a URL comparison guard, and hides the intro tooltip once stats have loaded
- CSS hides the stats frame when empty and hides the intro tooltip once `has-stats` is applied - _as soon as a user applies their first filter_

**Filter bug fixes**
- Source type filter was sending `"GW"`/`"SW"` — DB stores `"Groundwater"`/`"Surface Water"`
- Boundary type filter was querying `service_area_type` (EPA categorization) instead of `symbology_field` (map display type), and sending `"System"` instead of `"System Sourced"`
- Population size filter was sending numeric indices (`"1"`–`"5"`) instead of the stored range strings (`"<=500"`, `"501-3,300"`, etc.)
- All three fixed in both `#collectFilters` and `#restoreDomState` so URL-restored filter state is also correct
- These param/query patterns are now a 1:1 match with the legacy app, with an improvement on the Population size filter, where we make a query directly to that db column, instead of doing range filtering logic.

**Code quality**
- `build_summary` signature simplified — `systems_count` is now computed inside the aggregate rather than passed in as a separate argument
- Population category mappings extracted to module-level constants (`POP_CAT_MAP`, `POP_CLASS_MAP`) — one derived from the other, eliminating the duplicate inline objects
- Stats frame URL serialization uses `FilterState.toUrlParams()` (handles multi-value array params correctly) rather than `new URLSearchParams(obj)` which silently breaks arrays
- `symbology_field` added to `SORTABLE_COLUMNS`
- `pry-byebug` and `pry-rails` added to the `development, :test` gem group

**Docs and specs**
- `ARCHITECTURE.md` and `API.md` updated to reflect current routes, controllers, and endpoint behavior
- New `docs/TODO.md` capturing known gaps: frontend modernization (Hotwire/Tailwind), ETL field mapping validation, dev seed data filter coverage, controller refactoring, and observability
- Stats endpoint specs added with happy path, filter behavior, and edge cases (nil MHI, zero results, zero violations)
- `Filterable` specs extended to cover `symbology_field` and `pop_cat_5` range string values

## Notes for reviewers

The three filter bugs were pre-existing and unrelated to M7 — they were discovered during manual testing of this branch and fixed here since they were trivial and the same files were already open. The factory defaults for `pop_cat_5`, `symbology_field`, and `service_area_type` were also corrected as part of this.

`build_summary` calls `.left_joins(:demographic)` unconditionally. If `apply_filters` already joined demographics for a demographic filter, this produces a duplicate LEFT JOIN — PostgreSQL handles it cleanly on a non-nullable PK (no doubled rows, no error), which is why no guard was added.

---

### Manual

**Seed data reference** (VT + RI, 477 total systems):
| Filter | Expected systems count | Notes |
|---|---|---|
| :white_check_mark: No filter | 477 of 477 | |
| :white_check_mark: Has open violations | 23 of 477 | |
| :white_check_mark: Owner: Private only | 260 of 477 | |
| :white_check_mark: Owner: Local only | 206 of 477 | |
| :white_check_mark: Owner: State only | 7 of 477 | |
| :white_check_mark: Owner: Federal only | 2 of 477 | |
| :white_check_mark: Source: Groundwater | 371 of 477 | **bug fix** |
| :white_check_mark: Source: Surface Water | 106 of 477 | **bug fix** |
| :white_check_mark: Boundary: Modeled | 94 of 477 | **bug fix** |
| :white_check_mark: Boundary: System Sourced | 383 of 477 | **bug fix** |
| :white_check_mark: Population: smallest tier only (≤500) | 323 of 477 | **bug fix** |
| :white_check_mark: **Compound:** Source: Groundwater + Has open violations | 20 of 477 | customers: 8,893 · MHI: ~$73,262 · violations: 20 |
| :white_check_mark: **Compound:** Source: Surface Water + Owner: Local | 72 of 477 | |


app/controllers/public_water_systems_controller.rb
```diff
@@ -38,6 +38,13 @@ def show
     end
   end
 
+  def stats
+    scope = PublicWaterSystem.apply_filters(params)
+    unfiltered_total = PublicWaterSystem.count(:pwsid)
+    @summary = build_summary(scope).merge(unfiltered_total: unfiltered_total)
+    render layout: false
+  end
+
```
CSheesley, line 47: this action/route _(along with the export action/route)_ will be moved into their own controllers later in `TODO.md` to better follow the conventional Restful Routes pattern.

docs/API.md
CSheesley, line 1: Update this file to match the current state of our API endpoints

lukechanning (approved):
LGreatTM!

PR #9, opened 2026-04-16
[Feat] Remaining Missing Views for Dataset & Downloads (M12)(M14)

lukechanning:
## Summary

Populate the two remaining empty sidebar sections — Downloads (M12) and Datasets (M14) — by porting content from the legacy PHP app into ERB partials with a Stimulus controller for the datasets filter/sort UI.

## Changes

- **Bulk downloads page (M12):** national zip download link plus 54 state/territory links pointing at pre-packaged S3 zips, rendered as a grid matching the legacy layout
- **Datasets catalog (M14):** all 27 dataset cards ported from the legacy app, each with title, description, data source link, last-updated date, update frequency, and usage caveats
- **Datasets filter/sort:** Stimulus controller replacing the legacy jQuery + Isotope implementation — filter by data source (dropdown) and update frequency (button group), sort by date (newest/oldest), show-all reset
- **Security hardening:** `rel="noopener noreferrer"` on all `target="_blank"` links in both partials
- **Request specs:** 6 new specs covering download links (national, state, territory), dataset card rendering (content, metadata, count), and filter/sort control presence
- **Roadmap:** M12 checked off, M14 added and checked off

## Notes for reviewers

- No new controllers or routes — both sections are inline partials within the existing home page, consistent with how the nav section-switching pattern works (established in M4).
- The datasets partial is ~670 lines of editorial content (27 cards). The content is a faithful port from `deprecated/inc-datasets.php` — worth spot-checking a couple of cards against the source if you want to verify fidelity.
- The legacy CSS class `grid-conatiner` (typo) was preserved in the downloads partial because `water_tool.css` targets it with that exact misspelling. The `ds-update-frequecy` typo in the datasets partial was fixed since no CSS targets it.


CSheesley (approved):
Looks good to me - just two questions.

1) When we click on the one of those state download links, do we still get the same files exported in a `.zip`? For instance, this is what I got when I downloaded Colorado's:

Image

Is there an easy way to test this in code? Not sure if its worth adding, or already covered.

2) Given the FE changes here, are there any TODO items that we'll ant to add to the _(coming soon)_ TODO.md? If not, all good, just wanted to check.

PR #13, opened 2026-04-20
Refactor/make stats exports reports restful

CSheesley:
## Summary

Completes a two-part cleanup: moves three non-RESTful routes to conventional nested resources under `public_water_systems`, and wires up the previously stubbed CSV/GeoJSON export button to respect active filters.

## Changes

- **RESTful route refactoring** — three routes moved out of ad-hoc positions into proper nested resources under `public_water_systems`:
  - Stats: `GET /public_water_systems/stats` → unchanged path, but now a proper singular `resource :stats` in a collection block, handled by `PublicWaterSystems::StatsController`
  - Export: `GET /public_water_systems/export` → same path, now a proper singular `resource :export` in a collection block, handled by `PublicWaterSystems::ExportsController`
  - Report: `GET /reports/:pwsid` → `GET /public_water_systems/:pwsid/report`, nested as a member resource, handled by `PublicWaterSystems::ReportsController`

- **Export button**: Implemented the Stimulus export controller to read active filter state and trigger a filtered CSV or GeoJSON download. The `.csv`/`.geojson` radio selection is wired via Stimulus targets; the download respects whatever filters are currently applied to the table.

## Notes for reviewers

The export endpoint has always respected filter params server-side — this change just wires the frontend to pass them. No backend changes to the export logic.


lukechanning (approved):
Yeah, lookin' real good to me! I think this is a solid refactor — definitely merge whenever, solid!

PR #14, opened 2026-04-21
[Chore] deploy jobs etc

CSheesley:
**Note:** This is all the work of the great @lukechanning - I am just getting this merged in while he is on the other side of the world.

## Summary

Adds a GitHub Actions workflow for deploying to AWS ECS, alongside documentation for both the new ECS path and the existing Kamal deployment path. The workflow is opt-in via a repo variable so it can safely live in forks and community deployments without triggering failures.

## Changes

- **New CI/CD workflow** (`deploy-client-aws.yml`) that builds and pushes a Docker image to ECR on every push to `main` and on PRs, then force-redeploys the appropriate ECS service (production for `main`, preview for PRs)
  - Uses GitHub OIDC for keyless AWS authentication via an assumed IAM role
  - Tags images with both an immutable SHA-based tag and a mutable `latest`-style tag
  - Gated behind an `AWS_DEPLOY_ENABLED` repo variable — skipped (not failed) when unset
  - Concurrency group prevents multiple simultaneous deploys per target
- **README documentation** covering both deployment paths (Kamal and AWS ECS), including required repo variables/secrets and a note that AWS infrastructure provisioning is handled separately

## Notes for reviewers

The `AWS_DEPLOY_ENABLED` guard means this workflow is safe to merge without any repo variable set — it will simply appear as "skipped" in CI. Activation requires the AWS infrastructure to already exist and the repo variables in the README table to be configured.


CSheesley:
@lukechanning - From what I can gather, below is the original expected behavior of this script.
- I posted in the #drinking-water-tool-ext Slack Channel asking for some expected behavior information, that we should maybe try to match with deploys for this repo?
- Once we do button things up, we should remember to get this documented in the Repo, maybe in its own doc, which is referenced in the README?

####

  ---                                                                                                                                         
  1. Every merge to main is a production deploy.                                                                                        
  There's no gate beyond AWS_DEPLOY_ENABLED=true. Merge to main → Docker build → ECS update → live production. That's worth flagging          
  explicitly to whoever manages the repo.                                                                                                     
                                                                                                                                              
  ---             
  2. No staging environment.                                                                                                                  
  The PR preview slot is as close as it gets. There's no separate staging URL or environment in this setup.

  ---                                                                                                                                         
  3. PR deployments:
                                                                                                                                              
  a. URL — it's always https://water-data-tool-preview.policyinnovation.info. Fixed, not per-PR. Whichever PR was pushed most recently is what's live there. _(so no multiple pr deploys)_
   
  b. Infra teardown on merge — No. The preview service is always-on. When a PR is merged, the preview slot just keeps showing that image until the next PR overwrites :pr-latest. _(this is probably ok if we are ok with persisting a single pr deploy instance - but that could get troublesome unless PR branches are reviewed in a timely manner)_
                                                                                                                                              
  c. Branch naming convention — None required. The workflow triggers on any PR (opened, synchronize, reopened) regardless of branch name. _(again, fine if we only want a single pr deploy instance)_
   
  d. Data for PR preview — Maybe just a gap in the docs. It uses the water_data_tool_preview database on the same RDS instance (separate from production). But the docs say nothing about how data gets into that preview database. It would either need the ETL run against it manually, or it starts empty and schema-only until someone seeds it. _(maybe we can auto-seed any staging/pr deploys with a subset of data?)_

PR #15, opened 2026-04-22
[Fix] ETL Pipeline : single source and importing fixes

CSheesley:
## Summary

Removes the dependency on an externally-hosted JSON manifest file for ETL freshness detection, replacing it with direct S3 `Last-Modified` header checks per file. Also fixes two silent data bugs — a `cast_bool` error that was writing all three boolean indicator fields as `false` nationally, and a seed task gap that caused tribal water systems to never appear in local development.

## Changes

**ETL manifest → S3 HEAD requests**
- Replaced `ETL_MANIFEST_URL` (a manually-maintained JSON file hosted by CNT) with `ETL_SOURCE_URL`, a base HTTPS URL to the S3 folder of source data, managed by EPIC. The file list was already encoded in `FILE_IMPORTERS`; the ETL now issues one HTTP HEAD request per file at runtime and reads the `Last-Modified` header to determine freshness. No manifest file, no external dependency, no manual timestamp maintenance.
- `EtlImportJob` and the `etl:import` rake task are simplified — neither needs to fetch or pass a manifest URL; the importer handles it internally.
- The seed task's `download_data_files` method derives the file list from the same `FILE_IMPORTERS`/`FILE_EXTENSIONS` constants, keeping the two entry points in sync automatically.
- A single S3 bucket serves all environments (development, staging, production); `ETL_SOURCE_URL` controls which folder path each environment reads.

**`cast_bool` bug fix — boolean fields written as `false` nationally**
- `Etl::TypeCaster#cast_bool` compared against `"Y"` only, but the source CSV (`sdwis_viols.csv`) exclusively uses `"Yes"`/`"No"`. The result: `is_wholesaler`, `is_school_or_daycare`, and `is_grant_eligible` were written as `false` for every record regardless of actual source value. Fixed to accept both `"Y"` and `"YES"`. Note: the legacy PHP/Python app was unaffected because it wrote raw CSV strings to `TEXT` columns with no boolean casting — the Rails schema uses actual `boolean` columns, which is why the cast matters.
- Eliminated the duplicate `cast_bool` implementation in the seed task by replacing it with `extend Etl::TypeCaster` on `SeedImport`.

**Tribal systems missing from seed data**
- Tribal water systems use numeric EPA region codes as `pwsid` prefixes (e.g. `08...` for EPA Region 8) rather than two-letter state codes. The seed task's state-prefix filter (`pwsid.start_with?(state)`) never matched them, so `primacy_type = "Tribal"` was always 0 in local development.
- Added a second pass over `sdwis_viols.csv` after the state systems are loaded to collect tribal `pwsid`s. Their `epa_sabs` rows are then loaded and merged into `target_pwsids`. Steps 2–8 of the seed task were updated to use `CSV.foreach` + `target_pwsids` guard (rather than the state-prefix pre-filter) so tribal rows flow through all subsequent tables correctly.

**Code cleanup**
- Extracted duplicated HTTPS validation guard in `HttpFetcher` into a shared `validated_https_uri` private method used by both `fetch_url` and `head_url`.
- Extracted `SeedImport.build_pws_row` to eliminate an identical 13-field hash duplicated between the state and tribal seed passes.

**ETL schedule**
- Recurring job now runs at `12am America/New_York` (explicitly timezone-qualified) rather than an unqualified `6am` that defaulted to server-local time.

**Test coverage**
- New specs for `head_url` (HTTPS guard, SSL, response headers) and `build_file_entries` (URL construction, Last-Modified parsing, missing env var, missing header).
- New seed task specs covering the tribal path: verifies tribal systems seed alongside state systems, correct `stusps` from numeric prefix, and deduplication guard for overlap cases.
- Updated all existing ETL specs to remove `ETL_MANIFEST_URL` stubs and reflect the new HEAD-based flow.

## Notes for reviewers

- **Production boolean data:** The `cast_bool` fix means the next ETL run against production will correctly write `true` for wholesaler, school/daycare, and grant-eligible systems for the first time. This is the intended behavior — the legacy values were correct because they were never cast.
- **Confirm `ETL_SOURCE_URL` folder path with EPIC before go-live.** The `.env.example` value points to `test-staged/`. Staging and production deployments must be updated to point to the production data folder. This is flagged in `docs/TODO.md`.
- **Tribal systems in dev:** After re-seeding with `VT,RI,OH,CO,PR`, `PublicWaterSystem.where(primacy_type: "Tribal").count` should be > 0 (CO includes Ute tribal systems via EPA Region 8). The full `etl:import` was never affected by the tribal gap — only the state seed task.

### Before
**Local PWS data pulling from same source data bucket:**
Screenshot 2026-04-22 at 12 13 23 PM

**What the data looks like on Prod (CNT version), using same source bucket**
Screenshot 2026-04-22 at 12 13 46 PM

### After
**Local PWS data processed correctly, matches what in on the source data set**
Screenshot 2026-04-22 at 4 26 13 PM

**Local data now has records which match the previously affected queries**
Screenshot 2026-04-22 at 4 27 01 PM


app/jobs/etl_import_job.rb
```diff
@@ -3,12 +3,11 @@ class EtlImportJob < ApplicationJob
...
   #
-  #   EtlImportJob.perform_later                            # full import
-  #   EtlImportJob.perform_later(only: "epa_sabs")         # single table
+  #   EtlImportJob.perform_later                                 # full import
+  #   EtlImportJob.perform_later(only: "epa_sabs")               # single table
   #   EtlImportJob.perform_later(only: "epa_sabs", force: true)  # force re-import
   def perform(force: false, only: nil)
-    manifest_url = ENV.fetch("ETL_MANIFEST_URL") { raise "ETL_MANIFEST_URL is not configured" }
-    errors = Etl::Importer.new(manifest_url: manifest_url, force: force, only: only).call
+    errors = Etl::Importer.new(force: force, only: only).call
```
lukechanning, line 10: Thanks for cleaning this up, much appreciated!

app/services/etl/importer.rb
```diff
@@ -24,28 +24,40 @@ class Importer
       "pwsid_npdes_usts_rmps_imp" => Etl::Importers::PwsidNpdesUstsRmpsImp
     }.freeze
 
-    def initialize(manifest_url:, force: false, only: nil)
-      @manifest_url = manifest_url
+    FILE_EXTENSIONS = {
```
lukechanning, line 27: Niiiiice, great solve here!

db/seeds.rb
```diff
@@ -1,9 +1,13 @@
 # Production data is populated by the ETL pipeline (bin/rails etl:import).
 # This file seeds sample data for local development only.
 #
-# Seeds Vermont and Rhode Island — ~500 water systems total, enough to
-# exercise all features locally. To seed different states, edit the invoke
-# call below or run bin/rails 'db:seed:states[TX]' directly.
+# Seeds states for broad filter coverage in local development:
+#   VT + RI — small, fast to load, cover the common cases (~477 systems)
```
lukechanning, line 5: Damn, real nice stuff here. Love the specific state usage for clear local data coverage.

lib/tasks/seed_states.rake
```diff
@@ -124,34 +118,44 @@ namespace :db do
...
-          updated_at: Time.current
-        }
+        pws_rows << SeedImport.build_pws_row(row)
       end
 
       PublicWaterSystem.upsert_all(pws_rows, unique_by: :pwsid) if pws_rows.any?
       puts "  PublicWaterSystem:    #{pws_rows.size}"
 
+      # ---------------------------------------------------------------------------
+      # 1b. Tribal systems — collect from sdwis_viols.csv and add to target_pwsids
```
lukechanning, line 128: TIL — awesome catch, super cool. Thanks for building this in!

lukechanning (approved):
Man, truly awesome clean up of some serious tech debt with some groovy improvements. LGTM! 🚀

PR #16, opened 2026-04-23
[Feat] Map display dynamically changes based on selected Filters

CSheesley:
## Summary

Implements M10: when filters are applied, the map hides non-matching PWS polygons by fetching matching pwsids from a new `GET /map` endpoint and applying a Mapbox GL filter expression. Also fixes a bug where the "Tribal" ownership filter returned zero results due to a mismatch between the UI label and the DB value (`"Native American"`).

## Changes

- **Map ↔ filter integration (M10)** — `map_controller.js` now listens for `filters:changed`, fetches matching pwsids from `GET /map`, and applies a Mapbox GL `["in", "pwsid", ...]` filter expression to the `pws`, `pws_outline`, and `pws_points` layers. Resets to show all systems when no filters are active. Applied on page load so shared URLs with filter params render the correct map state immediately.
  - AbortController cancels in-flight requests on rapid Apply, preventing stale results
  - Guard against calling `setFilter` before map layers are added (race condition when `filters:changed` fires synchronously before `map.on("load")` completes)
  - `GET /map` added to `HomeController` — a lightweight endpoint that runs `apply_filters` and returns a bare `{ pwsids: [...] }` array. Sits alongside `GET /table` as a peer home-page data feed, not a PWS resource query

- **Owner type "Tribal" filter fix** — the Ownership filter checkbox labeled "Tribal" was sending `owner_type=Tribal` to the backend, but the EPA source data stores this value as `"Native American"`. Fixed the mapping in `filter_controller.js`. The UI label "Tribal" is intentional and matches the legacy app; the DB value is what changed.
  - Extracted the id↔value mapping to module-level constants (`OWNER_TYPE_MAP` / `OWNER_TYPE_ID_MAP`), eliminating a duplicate inline object in the URL restore path that had gotten out of sync

- **Dev seed creates `DataImport` record** — the seed task now writes a `DataImport` record on completion so the "Last updated on:" sidebar display works in development without needing a full ETL run

- **Frontend architecture decision doc** — added `docs/FRONTEND_DECISION.md` capturing the Hotwire vs. React SPA tradeoffs, backend infrastructure implications for each path, designer questions to resolve, and a current recommendation

## Notes for reviewers

The `GET /map` endpoint returns all pwsids when called with no filter params. This is intentional — the client always short-circuits with an early return before calling the endpoint when no filters are active. A server-side guard that returned `[]` for no-filter requests would change the endpoint's semantics incorrectly.

Known scale risk: at national dataset size, spreading tens of thousands of pwsids into a Mapbox GL filter expression may hit expression limits or cause latency. This is tracked in `docs/TODO.md`. The fix, if needed, is to pass filter params into the tile URL so `TilesController` applies filtering during MVT generation — eliminating the pwsid fetch entirely.


lukechanning (approved):
🚀

Curious to get some answers to FE questions, let's see!

PR #17, opened 2026-04-24
[FIX] Improve TileCache Generation and Homepage Performance

CSheesley:
## Summary

Performance and map quality improvements for the national-scale dataset. Replaces the blind tile cache warming strategy with a US bounding-box approach that covers two more zoom levels while using fewer total tile operations. Also removes a noisy centroid dot layer, tightens map zoom behavior, and adds territory navigation.

---

### Tile cache warming — smart z0–z8 with US bounding boxes

The old warm job generated tiles for every coordinate in the global grid for z0–z6, including empty ocean and land tiles. The new approach uses US region bounding boxes (continental US, AK, HI, PR, Guam+CNMI) to skip tiles with no PWS data, and extends warm coverage through z8.

- Warm coverage extended from z6 (large metro region zoom) to z8 (city/district zoom) — the level where users first see and interact with individual water utility polygons
- `MAX_WARM_ZOOM = 8` with bounding box filtering; z9+ generates on-demand (fast — small area, few polygons per tile)
- Fixed `perform_now` behavior: the previous implementation used `perform_later` per zoom level, which silently dropped jobs when run via `bin/rails runner` (no persistent worker); now runs all zoom levels synchronously inline
- Added live `$stdout` progress logging alongside `Rails.logger` so warm job progress is visible in the terminal

### `pws_points` layer removed (~20% tile cost reduction)

The centroid dot layer rendered green circles at low zoom. It was added in V2 but was not present in the legacy app, produced visual clutter over dense metro areas, and added a full extra layer (20%) to every tile generation and cache operation. Removed from `TileGenerator::LAYERS`, the SQL case block, and all map event handlers and filter calls.

### Map UX — zoom restrictions and territory navigation

- `renderWorldCopies: false` — prevents the map from rendering duplicate world copies when zoomed fully out
- `minZoom: 3` — restricts zoom-out to a North America view (previously allowed zooming out to see the whole world multiple times)
- Territory navigation buttons added for PR, GU, and MP with a visual divider separating them from state shortcuts. VI and AS have no PWS data and were intentionally excluded. All three territory zoom methods use the same two-step cinematic animation as AK and HI.

### ETL file importer progress logging

`FileImporter#call` now logs download progress, file size, elapsed time, and import completion to both `Rails.logger` and `$stdout` — making long-running ETL imports visible when invoked via `bin/rails runner`.

### Benchmarking tools

- `bin/benchmark` — safe to run at any time; measures filter queries, map pwsid pluck, stats aggregation, table page load, warm tile retrieval, and search. Never modifies the cache.
- `bin/benchmark-cold` — separate script for cold tile baseline measurements; requires explicit confirmation before wiping the tile cache and reminds you to re-run the warm job afterward.

---

### Smart z0–z8 vs old blind z0–z6

The most important result: **the new approach uses fewer tile operations than the old one while covering two more zoom levels.**

| Approach | Zoom coverage | Tile ops | Deepest warm zoom | Est. total time |
|----------|--------------|----------|-------------------|----------------|
| Old: blind z0–z6 | State-level view | ~21,844 | z6 — large metro region | ~12–19 min |
| **New: smart z0–z8** | **City/district view** | **~12,872** | **z8 — city/district** | **~32 min** |

The old approach warmed through z6 but generated 4,096 tile coordinates at that zoom — most of them empty ocean. The new approach uses 172 z6 coordinates (US-only) and extends through z8 with only 2,335 coordinates at that level. **96.5% reduction vs a naive full-grid z0–z9 approach** (3,213 coords vs 349,525).

### Why z8 is the sweet spot

| Zoom | Visual scale | Warmed? |
|------|-------------|---------|
| z6   | Large metro region | ✓ |
| z7   | Metro area | ✓ |
| z8   | City / district | ✓ ← warm floor |
| z9   | Neighborhood | on-demand |
| z10+ | Street-level | on-demand |

### Observed warm job timings (national dataset, ~44k systems)

| Zoom | Coords | Time | Per-coord |
|------|--------|------|-----------|
| z5 | 59 | 80.8s | 1.37s |
| z6 | 172 | 148.8s | 0.87s |
| z7 | 608 | 367.6s | 0.60s |
| z8 | 2,335 | 1,092.6s | 0.47s |
| **z0–z8 total** | **3,213** | **~1,940s** | **~32 min** |

Per-coord cost decreases at higher zooms — each tile covers a smaller area with fewer PWS geometries to aggregate.

z9 cold generation measured at 234ms. Pre-warming z9 would add ~53 minutes to the warm job for a quarter-second on-demand cost — not worth it.  

### Benchmark results — national dataset, 44,642 systems

**Tile generation (post-warm-job):**

| Tile | Cold | Warm | Speedup |
|------|------|------|---------|
| z=2 | 10,141ms | 47ms | **216×** |
| z=5 | 6,178ms | 26ms | **238×** |
| z=8 | 430ms | 4ms | **108×** |

The z=2 cold cost (10s) is not a query efficiency problem — EXPLAIN ANALYZE shows the bottleneck is `ST_AsMVT` aggregating ~24k geometries across 3 parallel workers. No missing index. This is why pre-warming exists: you pay the cost once, not per user. z=8 cold at 430ms confirms z9 on-demand is acceptable.

**Filter queries at national scale:**

| Query | Time |
|-------|------|
| No filter — count | 0ms |
| State filter (TX) | 2ms |
| Violations join | 30ms |
| EJ join (CEJST) | 35ms |
| Bounds filter (Chicago area) | 32ms |
| PWSID pluck — all 44k | 37ms |
| build_summary — no filter | 131ms |

---

## Notes for reviewers

- The warm job is destructive to the tile cache when run (it generates and upserts). Running `bin/benchmark-cold` also wipes the cache — the split into two scripts was specifically to prevent accidental cache destruction during routine benchmarking.
- `scratch/performance_work.md` contains the full session notes including the EXPLAIN ANALYZE breakdown, storage size measurements, and the reasoning behind all numerical constants.
- The `bin/benchmark` warm tile section will warn (not fail) if the cache is empty, prompting you to run the warm job first.

### DIFF
--- a/app/jobs/tile_cache_warm_job.rb
+++ b/app/jobs/tile_cache_warm_job.rb
@@ -1,37 +1,101 @@
 class TileCacheWarmJob < ApplicationJob
   queue_as :default
 
-  # Pre-generate tiles for zoom levels 0–6. Higher zooms are generated
-  # on demand — warming all z0–z8 tiles would be 87k+ coordinates.
-  # Each zoom level runs as a separate job so completed levels aren't
-  # re-done on retry and individual jobs stay short-lived.
-  MAX_WARM_ZOOM = 6
+  MAX_WARM_ZOOM = 8
 
-  def perform(zoom_level = nil)
-    if zoom_level
-      warm_zoom(zoom_level)
-    else
-      (0..MAX_WARM_ZOOM).each { |z| self.class.perform_later(z) }
+  # All [west, south, east, north] bounds must satisfy west < east (no antimeridian wrapping).
+  REGION_BOUNDS = [
+    [-125, 24, -66, 50],  # Continental US
+    [-180, 51, -130, 72],  # Alaska
+    [-161, 18, -154, 23],  # Hawaii
+    [-68, 17, -65, 19],  # Puerto Rico
+    [144, 13, 146, 16]  # Guam + CNMI
+  ].freeze
+
+  def perform
+    total_start = Time.current
+    total_coords = 0
+
+    log("[TileCacheWarm] Starting smart warm for z0-z#{MAX_WARM_ZOOM}")
+
+    (0..MAX_WARM_ZOOM).each do |z|
+      total_coords += warm_zoom(z)
     end
+
+    elapsed = (Time.current - total_start).round(1)
+    log("[TileCacheWarm] Complete - #{total_coords} coordinates warmed in #{elapsed}s")
   end
 
   private
 
   def warm_zoom(z)
-    tile_count = 0
-    grid_size = 2**z
+    started_at = Time.current
+    errors = 0
+    coords = tile_coordinates(z)
+    total = coords.size
 
-    grid_size.times do |x|
-      grid_size.times do |y|
-        TileGenerator.layers.each do |layer|
-          TileGenerator.generate_tile!(layer, z, x, y)
-        rescue => e
-          Rails.logger.error("[TileCacheWarm] #{layer}/z#{z}/#{x}/#{y}: #{e.class} — #{e.message}")
-        end
-        tile_count += 1
+    log("[TileCacheWarm] z#{z}: starting (#{total} coordinates, #{TileGenerator.layers.size} layers each)")
+
+    layers = TileGenerator.layers
+    report_interval = [total / 4, 1].max
+
+    coords.each_with_index do |(x, y), idx|
+      layers.each do |layer|
+        TileGenerator.generate_tile!(layer, z, x, y)
+      rescue => e
+        errors += 1
+        log("[TileCacheWarm] #{layer}/z#{z}/#{x}/#{y}: #{e.class} - #{e.message}", level: :error)
+      end
+
+      if ((idx + 1) % report_interval).zero?
+        pct = (((idx + 1).to_f / total) * 100).round
+        log("[TileCacheWarm] z#{z}: #{pct}% (#{idx + 1}/#{total} coordinates)")
       end
     end
 
-    Rails.logger.info("[TileCacheWarm] z#{z}: warmed #{tile_count} tile coordinate(s)")
+    elapsed = (Time.current - started_at).round(1)
+    error_note = (errors > 0) ? ", #{errors} error(s)" : ""
+    log("[TileCacheWarm] z#{z}: done in #{elapsed}s#{error_note}")
+
+    total
+  end
+
+  # Unions all US region bounding boxes and deduplicates overlapping tile coords.
+  def tile_coordinates(z)
+    seen = Set.new
+    coords = []
+
+    REGION_BOUNDS.each do |west, south, east, north|
+      x_min, x_max, y_min, y_max = bbox_to_tile_range(west, south, east, north, z)
+      (x_min..x_max).each do |x|
+        (y_min..y_max).each do |y|
+          coords << [x, y] if seen.add?([x, y])
+        end
+      end
+    end
+
+    coords
+  end
+
+  def bbox_to_tile_range(west, south, east, north, z)
+    n = 2**z
+    x_min = ((west + 180) / 360.0 * n).floor.clamp(0, n - 1)
+    x_max = ((east + 180) / 360.0 * n).floor.clamp(0, n - 1)
+    y_min = lat_to_tile_y(north, n)
+    y_max = lat_to_tile_y(south, n)
+    [x_min, x_max, y_min, y_max]
+  end
+
+  # Valid for lat in (-85.05, 85.05) — the Web Mercator bounds. Values outside this
+  # range produce NaN/Infinity that clamp silently to 0 or n-1.
+  def lat_to_tile_y(lat_deg, n)
+    lat_rad = lat_deg * Math::PI / 180
+    ((1 - Math.log(Math.tan(lat_rad) + 1.0 / Math.cos(lat_rad)) / Math::PI) / 2 * n).floor.clamp(0, n - 1)
+  end
+
+  def log(msg, level: :info)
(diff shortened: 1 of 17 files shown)


app/jobs/tile_cache_warm_job.rb
CSheesley, line 1: Logging in the terminal will look like this:
```
$ bin/rails runner "TileCacheWarmJob.perform_now"
[TileCacheWarm] Starting smart warm for z0-z8
[TileCacheWarm] z0: starting (1 coordinates, 4 layers each)
[TileCacheWarm] z0: 100% (1/1 coordinates)
[TileCacheWarm] z0: done in 37.9s
[TileCacheWarm] z1: starting (2 coordinates, 4 layers each)
[TileCacheWarm] z1: 50% (1/2 coordinates)
[TileCacheWarm] z1: 100% (2/2 coordinates)
[TileCacheWarm] z1: done in 38.1s
[TileCacheWarm] z2: starting (4 coordinates, 4 layers each)
[TileCacheWarm] z2: 25% (1/4 coordinates)
[TileCacheWarm] z2: 50% (2/4 coordinates)
[TileCacheWarm] z2: 75% (3/4 coordinates)
[TileCacheWarm] z2: 100% (4/4 coordinates)
[TileCacheWarm] z2: done in 39.5s
[TileCacheWarm] z3: starting (9 coordinates, 4 layers each)
[TileCacheWarm] z3: 22% (2/9 coordinates)
[TileCacheWarm] z3: 44% (4/9 coordinates)
[TileCacheWarm] z3: 67% (6/9 coordinates)
[TileCacheWarm] z3: 89% (8/9 coordinates)
[TileCacheWarm] z3: done in 43.6s
[TileCacheWarm] z4: starting (23 coordinates, 4 layers each)
[TileCacheWarm] z4: 22% (5/23 coordinates)
[TileCacheWarm] z4: 43% (10/23 coordinates)
[TileCacheWarm] z4: 65% (15/23 coordinates)
[TileCacheWarm] z4: 87% (20/23 coordinates)
[TileCacheWarm] z4: done in 52.1s
[TileCacheWarm] z5: starting (59 coordinates, 4 layers each)
[TileCacheWarm] z5: 24% (14/59 coordinates)
[TileCacheWarm] z5: 47% (28/59 coordinates)
[TileCacheWarm] z5: 71% (42/59 coordinates)
[TileCacheWarm] z5: 95% (56/59 coordinates)
[TileCacheWarm] z5: done in 74.3s
[TileCacheWarm] z6: starting (172 coordinates, 4 layers each)
[TileCacheWarm] z6: 25% (43/172 coordinates)
[TileCacheWarm] z6: 50% (86/172 coordinates)
[TileCacheWarm] z6: 75% (129/172 coordinates)
[TileCacheWarm] z6: 100% (172/172 coordinates)
[TileCacheWarm] z6: done in 143.1s
[TileCacheWarm] z7: starting (608 coordinates, 4 layers each)
[TileCacheWarm] z7: 25% (152/608 coordinates)
[TileCacheWarm] z7: 50% (304/608 coordinates)
[TileCacheWarm] z7: 75% (456/608 coordinates)
[TileCacheWarm] z7: 100% (608/608 coordinates)
[TileCacheWarm] z7: done in 343.3s
[TileCacheWarm] z8: starting (2335 coordinates, 4 layers each)
[TileCacheWarm] z8: 25% (583/2335 coordinates)
[TileCacheWarm] z8: 50% (1166/2335 coordinates)
[TileCacheWarm] z8: 75% (1749/2335 coordinates)
[TileCacheWarm] z8: 100% (2332/2335 coordinates)
[TileCacheWarm] z8: done in 1065.6s
[TileCacheWarm] Complete - 3213 coordinates warmed in 1837.4s
```

app/jobs/tile_cache_warm_job.rb
```diff
@@ -1,37 +1,101 @@
...
+  MAX_WARM_ZOOM = 8
+
+  # All [west, south, east, north] bounds must satisfy west < east (no antimeridian wrapping).
+  REGION_BOUNDS = [
+    [-125, 24, -66, 50],  # Continental US
+    [-180, 51, -130, 72],  # Alaska
+    [-161, 18, -154, 23],  # Hawaii
+    [-68, 17, -65, 19],  # Puerto Rico
+    [144, 13, 146, 16]  # Guam + CNMI
+  ].freeze
```
CSheesley, line 13: This narrows our scope so that we aren't taking time creating cached tiles for empty land or ocean areas, significantly reducing the number of tiles we create/store on the etl run

bin/benchmark
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env ruby
+# Benchmarks the key server-side query paths at national scale.
```
lukechanning, line 2: This is an extremely cool idea, should we fold this into the actual CI run we do locally as well? Come to think of it, I'm not sure we actually wired up a comprehensive `./bin/ci` for this, could be good. I like to run such things every time I am about to commit, it'd be cool to have benchmarking involved as well

lukechanning (approved):
Sure, seems wonderful to me! 🚀 Thanks so much for looking into this, I think the perf improvements and being more thoughtful here is a fantastic thing to do

bin/benchmark
CSheesley: Yeah it was kind of cool to see benchmarking - I'm not exactly sure how best to use it though. It could be cool for some sort of CI check to make sure we don't add some sort of major regression. That said - would or CI need to run a brand new ETL import call each time, or would it be able to reference a current cached version?

Maybe we hold off for now, but keep this in mind as we get things dialed in, and maybe identify opportunities for good benchmarking checks?

PR #130, opened 2026-04-28
[CHORE] Add  MAPBOX_STYLE_URL ENV Var - use EPICs account

CSheesley:
## Summary

This is a small PR which updates our Mapbox GL JS configuration to use the styling defined by EPIC in their Mapbox Studio account, which we set in the new `MAPBOX_STYLE_URL`. 

🚨 We will need to make sure that this value is set in AWS secrets, and referenced correctly in our Deploy Pipeline - _attn @lukechanning_

Two Workflows have been updated:
- **CI / TEST**: Both Mapbox ENV Vars have been set, using mock placeholder values
- **deploy-client-aws: Updated to fetch, set, and use the MAPBOX_STYLE_URL value _(to be set in AWS Secrets Manager)_

```
<meta name="mapbox-token" content="<%= ENV.fetch("MAPBOX_ACCESS_TOKEN") %>">
<meta name="mapbox-style" content="<%= ENV.fetch("MAPBOX_STYLE_URL") %>">
```

Both calls for Mapbox ENV vars have been updated to use `fetch` without a default value, so that if these ENV vars are not set, we will trigger a `Key Error` _(key not found)_, fail early, and get good feedback as to why things didn't work:

Ex:
Screenshot 2026-04-28 at 2 32 55 PM

While testing locally, this styling change was picked up and noticed in the UI: _(note the bold and black state names for example)_
Screenshot 2026-04-28 at 2 43 28 PM

### DIFF
--- a/.github/workflows/deploy-client-aws.yml
+++ b/.github/workflows/deploy-client-aws.yml
@@ -178,7 +178,7 @@ jobs:
             {"name":"SOLID_QUEUE_IN_PUMA", "value":"true"},
             {"name":"HTTP_PORT",           "value":"3000"},
             {"name":"TARGET_PORT",         "value":"3001"},
-            {"name":"ETL_SOURCE_URL",      "value":"https://tech-team-data.s3.us-east-1.amazonaws.com/national-dw-tool/test-staged"}
+            {"name":"ETL_SOURCE_URL",      "value":"${{ vars.ETL_SOURCE_URL }}"}
           ]'
 
           docker run --rm \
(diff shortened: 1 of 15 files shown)


.github/workflows/deploy-client-aws.yml
CSheesley, line 1: @lukechanning - are you able to set the AWS Secret for this new ENV Var accordingly, based on the changes in this workflow? I'm happy to make any changes if need be.

lukechanning: Yup, done! We'll need to set that secret still, but I have the ability to do that and it's in the infra now. I also need the live Mapbox access token, please! You can Slack that to me @CSheesley

lukechanning: Confirming the access token and style are set in AWS!

CSheesley: @lukechanning - thanks for getting that added, and ironing out all of the deploy stuff!

It looks like the `pr-deploy` job passed this time. Are you still using this branch for testing CI stuff against? If so, that is totally fine... if not, would you mind giving it a review/approval so that I can merge it in?

lukechanning (approved):
Let's 🚀 this one into prod so the ETL run in ~10 min can hopefully complete. Here we go!

PR #140, opened 2026-04-30
[Fix] Move to Streamable Geoms

lukechanning:
## Summary

Fixes a production OOM kill in the nightly ETL job caused by loading a 1 GB GeoJSON file entirely into memory — replaced with chunked HTTP streaming and SAX parsing so peak memory is O(batch_size), not O(file_size).

## Changes
- **ETL streaming fix for `epa_sabs_geoms.geojson`**
  - `HttpFetcher#stream_to_tempfile` streams the HTTP response body to disk in chunks via `Net::HTTP` block form — never buffers the full file in memory
  - `EpaSabsGeoms` importer rewritten to SAX-parse the tempfile one feature at a time using `Oj::Saj`, yielding features in 500-record batches to `import!` — peak memory stays constant regardless of file size
  - `oj` gem added as a dependency
  - Full spec coverage for both `stream_to_tempfile` and the SAX `FeatureHandler`

## Notes for reviewers

The `EpaSabsGeoms` importer now fully overrides `FileImporter#call` rather than delegating to `parse` — this is intentional since the SAX approach requires a different execution path (download → stream from disk) that doesn't fit the base class's synchronous parse model. The base class `import!` method is kept unchanged.

The `oj` gem's SAX interface (`Oj::Saj`) requires careful stack management in `FeatureHandler` — the spec exercises this directly against a real fixture file to catch any regression in the callback logic.

### DIFF
--- a/.github/workflows/deploy-client-aws.yml
+++ b/.github/workflows/deploy-client-aws.yml
@@ -178,7 +178,7 @@ jobs:
             {"name":"SOLID_QUEUE_IN_PUMA", "value":"true"},
             {"name":"HTTP_PORT",           "value":"3000"},
             {"name":"TARGET_PORT",         "value":"3001"},
-            {"name":"ETL_SOURCE_URL",      "value":"https://tech-team-data.s3.us-east-1.amazonaws.com/national-dw-tool/test-staged"}
+            {"name":"ETL_SOURCE_URL",      "value":"${{ vars.ETL_SOURCE_URL }}"}
           ]'
 
           docker run --rm \
--- a/.github/workflows/deploy-client-aws.yml
+++ b/.github/workflows/deploy-client-aws.yml
@@ -177,7 +177,8 @@ jobs:
             {"name":"RAILS_LOG_TO_STDOUT", "value":"1"},
             {"name":"SOLID_QUEUE_IN_PUMA", "value":"true"},
             {"name":"HTTP_PORT",           "value":"3000"},
-            {"name":"TARGET_PORT",         "value":"3001"}
+            {"name":"TARGET_PORT",         "value":"3001"},
+            {"name":"ETL_SOURCE_URL",      "value":"https://tech-team-data.s3.us-east-1.amazonaws.com/national-dw-tool/test-staged"}
           ]'
 
           docker run --rm \
--- a/app/services/etl/http_fetcher.rb
+++ b/app/services/etl/http_fetcher.rb
@@ -16,6 +16,27 @@ module Etl
       Net::HTTP.get(uri)
     end
 
+    # Streams the HTTPS response body to a Tempfile in chunks, never buffering
+    # the full response in memory. Returns the open, rewound Tempfile — caller
+    # is responsible for calling close! when done.
+    def stream_to_tempfile(url)
+      uri = validated_https_uri(url)
+      ext = File.extname(uri.path)
+      tmpfile = Tempfile.new(["etl_download", ext], binmode: true)
+
+      Net::HTTP.start(uri.host, uri.port, use_ssl: true) do |http|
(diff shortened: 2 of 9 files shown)


.github/workflows/deploy-client-aws.yml
```diff
@@ -177,7 +177,8 @@ jobs:
             {"name":"RAILS_LOG_TO_STDOUT", "value":"1"},
             {"name":"SOLID_QUEUE_IN_PUMA", "value":"true"},
             {"name":"HTTP_PORT",           "value":"3000"},
-            {"name":"TARGET_PORT",         "value":"3001"}
+            {"name":"TARGET_PORT",         "value":"3001"},
+            {"name":"ETL_SOURCE_URL",      "value":"https://tech-team-data.s3.us-east-1.amazonaws.com/national-dw-tool/test-staged"}
```
CSheesley: Is it ok that we are hardcoding this, as opposed to setting and referencing a Github ENV Var? I know its a public bucket, just wanted to check, since this might be the only spot we are referencing a hardcoded value.

app/services/etl/importers/epa_sabs_geoms.rb
```diff
@@ -27,7 +43,7 @@ def import!(rows)
                   INSERT INTO service_area_geometries (pwsid, geom, created_at, updated_at)
                   VALUES ($1, ST_GeomFromGeoJSON($2), NOW(), NOW())
                   ON CONFLICT (pwsid) DO UPDATE
-                    SET geom     = EXCLUDED.geom,
+                    SET geom       = EXCLUDED.geom,
                         updated_at = NOW()
                 SQL
                 "EpaSabsGeoms#import!",
```
CSheesley, line 49: This formatting seems odd, but StandardRB and the `bin/ci` check didn't seem to care... ship it.

app/services/etl/http_fetcher.rb
```diff
@@ -16,6 +16,27 @@ def fetch_url(url)
...
+          response.read_body { |chunk| tmpfile.write(chunk) }
+        end
+      end
+
+      tmpfile.rewind
+      tmpfile
+    rescue
+      tmpfile&.close!
+      raise
+    end
```
CSheesley, line 38: Clever

spec/services/etl/importers/epa_sabs_geoms_spec.rb
```diff
@@ -1,41 +1,128 @@
...
+
+  # Stubs stream_to_tempfile to return a real Tempfile backed by the fixture,
+  # avoiding any network call.
+  def stub_download(path = fixture_path)
+    tmpfile = Tempfile.new(["epa_sabs_geoms", ".geojson"])
+    tmpfile.write(File.read(path))
+    tmpfile.rewind
+    allow(importer).to receive(:stream_to_tempfile).and_return(tmpfile)
+    tmpfile
+  end
```
CSheesley, line 21: So the use of a temp file here is a bit of a divergence from patterns I am more familiar with - so I may be totally of base here - but is there a reason that we need to explicitly call this helper in each spec, as opposed to just using a single `before each` or `before all` call at the top of the spec?

CSheesley (approved):
Looks good to me - just left a couple of questions.

If possible, could we make sure that the `ETL.md` file is updated _(if necessary)_ to reflect these changes? I have noticed that some of these `/docs` files are falling out of accuracy with current behavior _(much of which was my doing)_ - I'm trying to do a better job of keeping these in sync when we can.

spec/services/etl/importers/epa_sabs_geoms_spec.rb
lukechanning: Mm, amazing catch. This is what happens when I'm going too fast! Should be fixed now

PR #145, opened 2026-05-06
[FEAT] FE Architecture Refactor - use Hotwire, Tailwinds CSS, etc.

CSheesley:
## Summary

This branch completes a full frontend architecture refactor — replacing DataTables and legacy jQuery patterns with Hotwire (Turbo Frames + Stimulus), Tailwind CSS, and ViewComponent — and delivers a substantially improved filter system with working histogram sliders, corrected filter logic, and a cleaner backend param pipeline.

### Frontend Architecture
- Replaced DataTables with a server-rendered Turbo Frame table with sorting, pagination (Pagy), null-last ordering, and a tiebreaker on name
- Refactored the monolithic filter controller into three focused Stimulus controllers: `filter` (state + apply), `filter-menu` (dropdown open/close), and `filter-layout` (responsive More menu overflow)
- Added `tooltip_controller`, `table_frame_controller`, and `dataset_card_controller`
- Migrated map/table UI, sidebar, datasets page, and system report sections to Tailwind; removed most of `water_tool.css`
- Introduced ViewComponent (`DatasetCardComponent`, `DetailSectionComponent`) with Lookbook previews and specs
- Consolidated all icons to inline SVGs with `fill=currentColor`; removed PNG assets

### Filter System
- Implemented per-sub-category histogram sliders for all 20 health violation sub-categories (10 per time window), each backed by a live `width_bucket` query with a shared in-memory cache
- Extracted `_slider_panel` partial and wired histogram sliders to the filter apply flow via hidden min/max inputs
- Fixed a bug where checking health sub-category filters at default slider positions sent no params — slider panels now seed hidden inputs with domain defaults when revealed
- Corrected filter logic to use **inclusive OR** across groups within a category (faceted search model): enabling Health 5yr OR Non-health 5yr broadens results rather than requiring both simultaneously; this applies across both time windows and between health and paperwork violation groups
- Extracted `FilterParams` PORO (`app/lib/filter_params.rb`) as the single source of truth for all permitted params — previously a growing inline list in the controller; also fixes several filter groups (EJ, funding, watershed hazard, trend, demographic) that were silently stripped by Strong Parameters

### Backend Fixes
- Fixed Strong Parameters: health sub-category `_min`/`_max` params were being stripped before reaching `Filterable`, making those filters non-functional
- Replaced bare integer interpolation in `ViolationsSummary#histogram_bins` with `connection.quote` for safe SQL parameter binding
[...]

PR #147, opened 2026-05-06
[Hotfix][5-7-26] Lookbook Gem Issues Causing Failed Builds

lukechanning:
## Summary

Fixes a production/staging crash introduced when ViewComponent previews were added in #145. The `lookbook` gem is development-only, but `app/components/previews/` was being eager-loaded by Zeitwerk in all environments, causing a `NameError: uninitialized constant Lookbook` on boot and a 503.

## Changes

- Tell Zeitwerk to ignore `app/components/previews/` unconditionally in `config/application.rb` — Lookbook discovers preview files by scanning the filesystem itself, not through the autoloader, so development behaviour is unchanged
- Guard the ViewComponent previews path registration in the initializer to development-only, since it's irrelevant in other environments

## Notes for reviewers

The `Rails.autoloaders.main.ignore` call must live in `config/application.rb`, not an initializer — the autoloader is already set up by the time initializers run. Confirmed fix with a `RAILS_ENV=production rails runner` boot check before committing.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,6 +28,11 @@ module WaterDataTool
     # Common ones are `templates`, `generators`, or `middleware`, for example.
     config.autoload_lib(ignore: %w[assets tasks])
 
+    # Lookbook preview files inherit from Lookbook::Preview, a development-only gem.
+    # Zeitwerk must ignore this directory in all environments; Lookbook discovers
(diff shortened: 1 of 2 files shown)


CSheesley (approved):
🤦 Thanks for cleaning up my mess! I'm not super familiar with LookBook, nor do I know how helpful it will ultimately be for us - we'll see!

Image

PR #153, opened 2026-05-11
[FEAT] Add Remaining Filters in UI

CSheesley:
## Summary

Completes the remaining filter groups (Population, Financial, Funding, Environmental) and introduces a `FilterRegistry` as a single source of truth for all filter configuration. This replaces scattered filter definitions with a declarative `config/filters.yml` and wires up histograms across all relevant models.

## Changes

- **`FilterRegistry` + `config/filters.yml`** — Centralizes all filter URL parameter keys, range column definitions, and histogram field config in one YAML-backed registry. Eliminates duplicated filter lists across `Filterable`, the histogram controller, and the Stimulus controller.

- **`Histogrammable` concern** — New shared concern that provides histogram query support; wired to `Demographic`, `EnvironmentalJustice`, `FundingSummary`, `TrendDatum`, `ViolationsSummary`, and `WatershedHazard`.

- **`RangeFilterItemComponent`** — New ViewComponent for range filter UI items, handling both nested (toggle-to-expand) and flat slider variants with tooltip and accessibility support.

- **Remaining filter groups** — Population, Financial, Funding, and Environmental filter sections added to `_filter_menus.html.erb`, with all corresponding histogram sliders and tooltip copy.

- **Histograms controller** — Extended to route histogram requests to the correct model per field, using `FilterRegistry.histogram_field_config`.

- **`FilterParams` relocated** — Moved from `app/lib/` to `app/filters/` to match the new filter-layer conventions.

- **Filter Stimulus controller** — Embeds the `FilterRegistry` client payload as inline JSON (via a server-rendered partial) so the JS layer reads from the same config without duplication.

- **`config/tooltips.yml`** — Tooltip copy added for all new filter fields.

## Notes for reviewers

The `FilterRegistry` is the main architectural addition here — it's a memoized class that loads `config/filters.yml` at boot and exposes typed accessors used by `Filterable`, `Histogrammable`, the histogram controller, and the Stimulus filter controller. `FilterRegistry.reload!` is available in tests to reset memoized state.

The histogram endpoint now fans out to multiple models per request (one field → one model). If a field isn't in the registry, it returns a 400, which is a change from the previous silent pass-through.


config/application.rb
```diff
@@ -28,10 +28,13 @@ class Application < Rails::Application
...
+    # Custom app/filters/ is not on Rails' default autoload list
+    config.autoload_paths << Rails.root.join("app/filters")
+
     # Lookbook preview files inherit from Lookbook::Preview, a development-only gem.
-    # Zeitwerk must ignore this directory in all environments; Lookbook discovers
-    # previews by scanning the filesystem itself, not through the autoloader.
-    Rails.autoloaders.main.ignore(Rails.root.join("app/components/previews"))
+    # Ignore in non-development so Zeitwerk doesn't try to load Lookbook::Preview where the gem is unavailable.
+    # In development, Zeitwerk loads them normally so Lookbook can discover the classes.
+    Rails.autoloaders.main.ignore(Rails.root.join("app/components/previews")) unless Rails.env.development?
```
lukechanning, line 37: Hey uh, great catch 😆 My bad!

lukechanning (approved):
This is such a cool way to do this, man. I definitely dig it — nice to move something that was hard code (not necessarily "hardcoded") into something more configurable via `filters.yml`.

I'm into it, let's 🚢 it!

config/application.rb
CSheesley: All good - I didn't know and I reviewed your stuff. Its also a pretty unused feature as of yet, but looks like this:

Screenshot 2026-05-12 at 10 25 38 AM

PR #160, opened 2026-05-14
[Feat][Refactor] Adjust County Handling in Pipeline + General Improvements

lukechanning:
## Summary

Replaces the spatial-intersection approach to computing county associations with a pre-computed CSV importer, and extracts the cartographic boundary loader from a one-off Rake task into a proper service class. Together these make the ETL pipeline cheaper to run, easier to test, and more consistent in how each data source is owned.

## Changes

- **New `sabs_pwsid_county` CSV importer** — populates `public_water_systems.counties` from an authoritative CSV rather than inferring county membership from geometry intersections. Handles multi-county rows, multi-row aggregation per PWSID, deduplication, and alphabetical sorting. Pre-filters to known PWS records before upserting to prevent ghost record creation.

- **Removed `build_county_associations`** — the PostGIS spatial join that previously computed counties from `cartographic_counties` is gone. The CSV is now the authoritative source; counties are no longer a derived output of geometry imports.

- **`CartographicBoundaries` service** — the `cartographic:load` Rake task and its inline `CartographicLoader` module have been replaced with a proper `CartographicBoundaries` service class. Improvements over the Rake task: database password is passed via `PGPASSWORD` env var instead of embedded in the connection string (hidden from `ps aux`), each layer's staging→target swap is wrapped in a transaction, and extracted shapefiles are cleaned up after each successful load (zips retained for re-run caching).

- **`CartographicBoundaries.loaded?` idempotency guard** — Census boundaries change at most once a year; `PostImportSteps` now skips the shapefile reload if all three boundary tables are already populated.

- **`PostImportSteps.call` takes `imported_files:` array** — instead of the caller tracking separate `any_imported` and `geometry_imported` booleans and calling `bust_tile_cache`, `PostImportSteps.call`, and `TileCacheWarmJob` separately, `Etl::Importer` now passes the list of imported file keys and `PostImportSteps` decides internally what to run. Tile cache busting and the warm job fire for any import; geometry-dependent steps (centroid generation, state code assignment, spatial index rebuild, place crosswalks) are gated on `epa_sabs_geoms` being in the list.

## Notes for reviewers

The removal of `build_county_associations` means counties are no longer self-healing from geometry — if a PWSID is absent from the CSV, its existing `counties` value is preserved rather than cleared or recomputed. This is intentional (conservative by design) and matches how every other CSV-driven column works in this pipeline. Worth confirming with stakeholders that the CSV source is considered authoritative going forward.

Cross-layer atomicity in `CartographicBoundaries` is per-layer, not global — if the second shapefile fails, the first is already committed. A re-run corrects partial state. This matches the previous Rake task's behaviour.


app/services/etl/importers/sabs_pwsid_county.rb
```diff
@@ -0,0 +1,36 @@
+require "csv"
+
+module Etl
+  module Importers
+    class SabsPwsidCounty < Etl::FileImporter
+      def parse(content)
+        grouped = Hash.new { |h, k| h[k] = [] }
+
+        CSV.parse(content, headers: true) do |row|
+          # Skip blank rows. If a PWSID's every row is blank it won't appear
```
CSheesley: Is this just a typo - `If a PWSID's every row is blank it won't appear`? Is that supposed to say 'counties served'?

app/services/etl/importers/sabs_pwsid_county.rb
```diff
@@ -0,0 +1,36 @@
...
+        end
+      end
+
+      def import!(rows)
+        return if rows.empty?
+
+        known_pwsids = PublicWaterSystem.where(pwsid: rows.map { |r| r[:pwsid] }).pluck(:pwsid).to_set
+        valid_rows = rows.select { |r| known_pwsids.include?(r[:pwsid]) }
+
+        PublicWaterSystem.upsert_all(valid_rows, unique_by: :pwsid, update_only: [:counties]) if valid_rows.any?
```
CSheesley, line 32: This is just a more general questions for the ETL pipeline - are our importers going to run one after another _(one at a time)_? If so then these upsert calls should be good.

app/services/etl/post_import_steps.rb
```diff
@@ -4,11 +4,26 @@ module Etl
...
 
-    def call
+    def call(imported_files:)
+      # imported_files is an array of successfully-imported file keys, e.g. ["epa_sabs", "epa_sabs_geoms"]
+      return if imported_files.empty?
+
+      # If any files were imported, bust the tile cache to ensure all tiles are regenerated with fresh data.
+      # This is a broad hammer but ensures no stale tiles are served after any import.
+      bust_tile_cache
+      TileCacheWarmJob.perform_later
```
CSheesley: Just to confirm, we do want to run these two calls regardless if the data is geographically focused, is that right? If so - can you remind me why?

At the very least - are we still checking against file update at timestamps to gate processing any unchanged files?

app/services/cartographic_boundaries.rb
```diff
@@ -0,0 +1,143 @@
...
+  def self.load
+    new.load
+  end
+
+  # Returns true when all three boundary tables contain data. Used by callers
+  # to skip a reload when boundaries are already in place — Census geometries
+  # change at most once a year, so reloading on every ETL run is unnecessary.
+  def self.loaded?
+    CartographicState.exists? && CartographicCounty.exists? && CartographicPlace.exists?
+  end
```
CSheesley, line 41: Will these already exist when the ETL runs the next time that census data is updated? Or do we have some sort of logic that clears these out IF we are getting new census data files?

CSheesley (commented):
I think this looks pretty good to me - impressive work!

I did leave a few questions that I was curious about, and hoping to solidify my understanding of things.

Once we are about to merge things in, can we just make sure that the `ETL.md` file is updated accordingly?

app/services/etl/importers/sabs_pwsid_county.rb
lukechanning: Updated the code comment here to clarify, great catch. It's a partial truth, but one that's more obvious now.

app/services/etl/importers/sabs_pwsid_county.rb
lukechanning: To my understanding, the ETP pipeline calls the importers one at a time, yes absolutely!

app/services/etl/post_import_steps.rb
lukechanning: Code commented updated here as well!

app/services/cartographic_boundaries.rb
lukechanning: Okay, so I think this is a really good clarification — I think a _bad_ approach here is requiring some kind of manual truncation of whatever. 

What if instead we do either a Rails cache fingerprint or a DB-backed url fingerprint we can check against? Something like: 

1. **Rails.cache**
- On load, write a cache key derived from a hash of the LAYERS URLs; loaded? checks row presence AND that cache key exists
- Zero migrations; downside is a cold cache (deploy, flush) triggers a ~30s reload even if nothing changed

2. **DB-backed fingerprint**
- Store the LAYERS URL hash in a new column (e.g. on data_imports) when boundaries load; loaded? compares stored hash to current
- Always accurate, never a false-positive reload; requires a migration

lukechanning: @CSheesley — do we wanna try to address this in this work (assuming we pick a solid direction) or do we wanna push that to a new issue?

app/services/etl/post_import_steps.rb
CSheesley: Awesome, thanks for the info add, and reminding me why we still do that! I didn't realize that the tile cache held more info that geometry data.

app/services/cartographic_boundaries.rb
CSheesley: I'm not sure of the best way to handle this in the longterm - whether it be in code or as part of a process that the data team understands/owns.

Presumably, when new census data is update, that results in new files/urls - so maybe your thought about looking at LAYERS URL make sense?

Maybe we ship this as is... but created a ticket to sync up and talk about this? I think when I first pointed out the hard coded urls here, the data team was cool with this approach for now - but this is the second time we have paused to talk/look at this file. Maybe we just need to solidify our understanding of how/when these files update?

CSheesley (approved):
Approved! Thanks for all of the answers to my questions, and helping me remember how the inner workings of the ETL operate.

As for handling the Census data updates and the Cartographic<*> models - I left a comment there about maybe taking some time to strategize around that piece, but maybe that can happen outside of this work (with a diff ticket) unless we feel like we can lock down a solid solution here. Your call!

PR #164, opened 2026-05-16
[FEAT] Make Section Panels float as Cards over persistent Map Background

CSheesley:
## Summary

Refactors the main layout so the map canvas stays visible as a persistent background in all views. Table, datasets, downloads, and documentation now appear as floating cards on top of the map rather than replacing it. Also removes a large block of float-based legacy CSS that was propping up the old full-takeover layout, consolidates stylesheet rules into `application.css`, and replaces CSS-class-driven state with data-attribute state on the datasets page.

## Changes

- **Persistent map background** — `#container-map` is no longer hidden when switching views. A `.table-mode` / `.section-mode` class on the container drives CSS that hides interactive overlays (filter bar, geocoder, region nav) when a section card is active. The map canvas remains visible as a decorative background in all views.

- **Section cards** — Table, datasets, downloads, and documentation panels are now positioned cards (`absolute`, `rounded-xl`, `shadow-lg`) that float over the map. `sidebar_controller` shifts their left edge when the sidebar collapses or expands.

- **Table view filter bar** — In table mode the filter bar is pinned inside the table card so it acts as the card's header, rather than floating freely over the map.

- **Datasets page: data-attribute state** — `dataset_card_controller` now toggles `data-collapsed="true/false"` instead of adding/removing a CSS class; clip and expand styles are driven by Tailwind `data-[collapsed=true]:` variants in the markup. Filter and sort buttons use `data-active="true/false"` with the same approach, replacing the global `.is-checked` utility class.

- **Downloads page: Tailwind migration** — Replaced the old float-based grid and `.container-section-inner` layout with a responsive CSS Grid and Tailwind utilities.

- **Legacy CSS removal** — Deleted ~200 lines from `water_tool.css`: `.clear`/`.clearfix`, `#container-map` and `#map` positioning, `.container-section-inner`, stats bar/intro tooltip rules, downloads grid, table scrollbar rules, and related mobile overrides — all superseded by Tailwind utilities or moved to `application.css`.

- **`application.css` consolidation** — Section-mode overlay rules, WebKit scrollbar styles for both filter menus and the data table, and Firefox scrollbar rules are now centralized here. Removed the equivalent inline Tailwind arbitrary properties from `FilterMenuPanelComponent`.

## Notes for reviewers

The section-mode CSS state machine (`.table-mode` / `.section-mode` on `#container-map`) is the load-bearing architecture change. If `nav_controller.js` fails to set those classes correctly — or if a Turbo navigation restores a cached snapshot mid-view — the map overlays may not hide/restore as expected. Worth clicking through all five nav items (Map, Table, Datasets, Downloads, Documentation) and back-navigating with the browser to confirm state resets cleanly.

---
Screenshot 2026-05-16 at 4 24 11 PM

Screenshot 2026-05-16 at 4 24 40 PM


app/javascript/controllers/sidebar_controller.js
```diff
@@ -3,7 +3,7 @@ import { Controller } from "@hotwired/stimulus"
 const FULL_WIDTH = 250
 const COLLAPSED_WIDTH = 80
 const SIDEBAR_LEFT = 16      // matches left-4 (16px)
-const CONTROLS_GAP = 8       // gap between sidebar right edge and map controls
+const CONTROLS_GAP = 16      // gap between sidebar right edge and content — matches SIDEBAR_LEFT so all three outer gaps are equal
```
lukechanning, line 6: I missed that we were running these as set constants, that's actually awesome — super stable and solid pattern, great stuff!

lukechanning (commented):
Looking SO sick, thanks for doing this!

app/javascript/controllers/sidebar_controller.js
CSheesley: Thats the hope! This might change once I revisit things with mobile in mind (we'll see!)

Issue #180, opened 2026-05-25
[FEAT]: Columns Drag and Drop - add drag and drop to data table


CSheesley:
### Team / Area

Frontend

### Requested By

csheesley

### Overview

As a follow along to #117 , this PR will segment and track the new Feature work of adding Drag and Drop ability to to the Data Table's columns, via drag and drop interactions inside of the 'Manage Columns' modal.

### Context

Work has already started _(on #117)_ which will allow a user to choose which columns to hide/show when viewing the data table. That work will include a 'manage columns' modal which lists columns, and will include place holder 'drag and drop' icons - which will not be wired up yet.

### Success Criteria

- [ ] A user is able to successfully reorder columns via the drag and drop feature
- [ ] The rendered table respects the ordering of the user's drag and drop selections
- [ ] The/A reset button will reset column ordering to its default state.

### Dependencies / Blockers

- Depends on #117

PR #181, opened 2026-05-25
[FEAT] Add ability for user to Manage Columns on Data Table

### LINKED ISSUE #117: Update the display of the table view to more closely match the design specified in the design system including the ability to select individual rows
The table view of the data shown on the map should be updated to match the specified design, including the ability to select individual data rows so that people can create custom data packages for download. Elements that are shown across table and map view should remain consistently placed on the screen. The table view should scroll horizontally and vertically. There was pagination added to this screen for performance reasons that is not currently reflected in the design. 

Image

Definition of done:
- [x] Table view reflects the provided design
- [x] People are able to select/deselect rows of interest (plus select all) with selected rows displayed as shown
- [x] People are able to horizontally and vertically scroll the data
- [x] Components are more customized as shown
- [x] Table view is more performant than current implementation

CSheesley:
🚨 LARGE PR ALERT...

# Data Table — Row Selection, Streaming Export, Column Groups & URL Persistence

### Ticket(s)
- #117
- #205
- #206 _(partially - POST requests addressed here)_
- _Future drag-and-drop column reordering will happen on #180_

---

## Summary

- The ~70 hardcoded column definitions that lived in ERB were replaced with a YAML-driven registry. All column metadata (label, sort key, format, size, visibility flags, SQL expression) now lives in `config/columns.yml`. `ColumnRegistry` and `TableColumn` back it.
- Row selection uses an **Inversion of Selection** state model — all rows selected by default, individual unchecks tracked as exclusions. Selections survive page navigation and column changes; filter changes clear them.
- CSV and GeoJSON export are now fully streaming (no AR objects instantiated). CSV preserves the user's current sort order and visible column set. GeoJSON is full-fidelity via cursor-based batching.
- The Manage Columns panel now groups columns by category with collapse/expand, select-all-in-category, and indeterminate header states.
- `cols=`, `sort=`, and `direction=` are now preserved across filter changes and table navigations — column and sort choices survive the full session.

---

#### Column Registry (data-driven rendering)

- The ~70 hardcoded column definitions that lived in ERB were replaced with a YAML-driven registry _(following the common pattern in the app)_.
- All column metadata (label, sort key, format, size, association, visibility flags) now lives in `config/columns.yml`.
- Two new classes back it: `ColumnRegistry` (loads/memoizes from YAML) and `TableColumn` (Data.define value object).
- The table `<thead>` and `<tbody>` are now registry-driven loops. `render_table_cell`, `cell_value`, and `format_cell_value` helpers in `HomeHelper` handle all cell rendering.

**Why it matters:** eliminates ~400 lines of duplicated ERB, makes adding/removing columns a YAML edit, and is the foundation for dynamic column visibility and ordering.

---

#### Sortable Column Headers

- Extracted column header rendering into `UI::TableHeaderComponent`. Handles sort URL generation, direction toggling, ARIA labels, and size variants.
- `SORTABLE_COLUMNS` and `TABLE_JOINS` moved from a controller constant into `config/filters.yml` via `FilterRegistry`.

---

#### Row Selection + Scoped Export

- Checkbox column added as the first sticky column in every row.
- **State model — Inversion of Selection pattern** (used by Salesforce, AWS Console, HubSpot for large paginated datasets):
  - Default: **All mode** — `excluded = {}` means every row is implicitly checked. Unchecking a row adds its ID to the `excluded` Set; re-checking removes it.
  - **None mode** — entered via Deselect All only. `included = {}` means nothing is checked; checking a row adds it to the `included` Set.
  - `isAllChecked()` and `isAllMode()` are derived predicates, not state. Source of truth is `mode` + `excluded` Set + `included` Set.
  - Sets are used (not arrays) — membership testing is O(1), and uniqueness is guaranteed by construction.
- **Select All / Deselect All** buttons sit above the table (outside the Turbo Frame, so they persist across navigations).
- **Export badge** on the Export button shows `All`, a live numeric count, or `0` depending on selection state.
- **Export button disabled state** when nothing is selected: background swaps to gray, `cursor-not-allowed`, `aria-disabled="true"`, and a native `title` attribute (consistent with `title="Opens in new tab"` on external links).
- Individual row unchecks persist across page navigation.
- Selection clears on filter changes; column visibility changes do not clear it.

**Three export paths (all via POST):**

| Selection state | What is sent | Server behaviour |
|---|---|---|
| All mode, nothing excluded | Filter params + search | `apply_filters` + `apply_search` |
| All mode, some excluded | Filter params + search + `exclude_pwsids[]` | Above + `where.not(pwsid: ...)` |
| None mode, rows checked | `pwsids[]` | `where(pwsid: ids)` |
| None mode, nothing checked | no-op | early return, no submission |

Export always uses POST via a dynamically built `<form>` (with CSRF token) submitted and immediately removed. Sort, direction, and search are read from a `#table-query-state` sr-only span rendered inside the Turbo Frame — the frame always reflects live state regardless of whether those params appear in `window.location`.

---

#### Streaming CSV & GeoJSON Export

Both export formats now use the Rack body streaming pattern (`self.response_body = enumerator`). No AR objects are instantiated on either path. Neither sends `Content-Length` — the browser shows a live byte count as data arrives.

**CSV — two-phase approach (sort preserved):**
- **Phase 1 — sorted ID pluck:** `@scope.pluck(:pwsid)` captures the user's requested sort order (~2–4 MB for the full dataset). No AR objects.
- **Phase 2 — batched raw SQL:** IDs are sliced into 1,000-element chunks. Each chunk is fetched via a single raw SQL query with all LEFT JOINs inline. Rows within each batch are re-ordered by the phase-1 sequence before streaming.
- Export columns are determined by `ColumnRegistry.csv_columns(keys: visible_keys)` — only the user's currently visible columns are exported. GeoJSON remains full-fidelity (all columns).

**GeoJSON — cursor-based batching:**
- `WHERE pws.pwsid > last_seen_id ORDER BY pws.pwsid LIMIT 1000` keeps memory flat regardless of result size.
- `ST_AsGeoJSON` + `jsonb_build_object` build each Feature inline in SQL. The 100-arg PostgreSQL limit is handled by slicing column pairs at 49 and merging chunks with `||`.
- `ASSOCIATION_JOINS` is a shared frozen constant used by both CSV and GeoJSON queries.

**What was removed:**
- `with_details` (an 8-association `includes(...)` call) was removed from the CSV export path — it materialised every AR object and all its associations into the Ruby heap before writing a single byte.
- `CSV_HEADERS` and `CSV_COLUMNS` (two parallel 71-entry arrays requiring manual sync) were replaced by a single `ColumnRegistry`-driven source of truth.
- The GET export route and `show` action were removed. GET breaks silently for large payloads (URL length limits) —  the exact problem POST was introduced to solve. Route is now `resource :export, only: :create`.
- `send_data` replaced by direct `self.response_body` assignment.

---

#### Manage Columns: Category Grouping

- `CategoryDef = Data.define(:key, :label)` value object added.
- `config/columns.yml` gains a top-level `categories:` block (ordered list) and a `category:` field on each column entry. Categories: `utility_details`, `violations`, `demographics`, `environmental_justice`, `funding`, `watershed_hazards`.
- `ColumnRegistry.categories` — memoized, returns `Array<CategoryDef>` in YAML-defined order.
- **3 ViewComponents** extracted for the manage-columns panel:
  - `ManageColumns::PinnedRowComponent` — always-visible disabled row.
  - `ManageColumns::CategoryHeaderRowComponent` — category label + collapse/expand chevron + toggle-all checkbox.
  - `ManageColumns::ColumnRowComponent` — checkbox row with a drag handle icon; `indented: true` shifts padding for category children. Drag-and-drop is not yet wired (deferred to #180).
- `manage_columns_controller.js` additions:
  - `toggleCategoryCollapse` — toggles `aria-expanded`, hides/shows child `<ul>`, rotates chevron.
  - `toggleCategory` — checks/unchecks all children; uncheck collapses, check expands.
  - `syncCategoryState` — updates the category header checkbox and indeterminate state when an individual child changes. Header is checked if ALL children checked, indeterminate if SOME, unchecked if none.
- `config/environments/development.rb` — `config.watchable_files` extended with `columns.yml` and `filters.yml` so Zeitwerk reloads on config changes (previously required a server restart).

---

#### URL State Persistence

`cols=`, `sort=`, and `direction=` are now preserved across filter changes and table navigations. Previously, applying a filter clobbered column and sort choices.

**Root cause:** `filter_controller.js` `#syncToUrl()` replaced the entire URL search string with filter params only, discarding display params. `#visitTableFrame()` built the Turbo visit URL from filter state only.

**Three changes:**
1. `filter_controller.js` `#syncToUrl` — preserves `cols`, `sort`, `direction` when replacing filter params in the URL.
2. `filter_controller.js` `#visitTableFrame` — now visits `window.location.href` directly (URL is already complete state after change 1).
3. `manage_columns_controller.js` — form-submission approach replaced with `Turbo.visit`. Removes the `colsInput` hidden target and form submit wiring; `serializeCols` and `reset` share a `#applyAndVisit(keys)` private helper that updates the URL and visits the Turbo Frame.

**Reset scoping:**
- **Reset All** (filter panel) — resets filter state only; `cols`, `sort`, and `direction` remain.
- **Reset** (Manage Columns) — deletes `cols` only; filters and sort remain.

---

### Bug Fixes & Housekeeping

- `open_health_viol` attribute changed from `:string` to `:boolean` — migration, casting, and spec fixes.
- `ExportsController` — strong params enforced via `export_params` private method. `build_export_scope` / `filtered_scope` private methods replace nested if/else scope building. Action is now two lines.
- `PublicWaterSystemsController` — simplified to JSON-only; orphaned HTML show view deleted.
  - _Open question: Do we want to keep/maintain these currently unused API-style endpoints?_
- `FilterParams` — consistency fix from earlier review triage.
- `UI::TableHeaderComponent` — `check?` method removed. The check column has `label: nil` and `sort: nil`, so `sortable?` is already false and `<%= nil %>` renders nothing; no special case needed.
- Export button color restored — a rebase accidentally changed `bg-[#67a25e]` (green) to `bg-brand-primary` (blue). Fixed in both `index.html.erb` and the JS disabled-state toggle.
- Filter panel toggle arrows aligned to match columns panel convention (`↓` open, `→` closed): `rotate-180` changed to `-rotate-90` in `#setToggleArrow`, reset block, and initial classes in `_filter_menus.html.erb` and `group_range_component.html.erb`.
- Filter panel parent sync (`syncParentFromSubcat`) — moved to after the per-subcat `forEach` loop so it reads actual checked state instead of always forcing `checked = true`.
- `home_controller.rb` `params[:cols]` check changed from `.blank?` to `.nil?` to correctly distinguish absent (`nil` → all columns), empty (`""` → pinned only), and explicit (`"pwsid,stusps"` → named set).

---

## Notes

- The `cols=` URL param uses a comma-separated string (not array params) intentionally — the order of values encodes column order, positioning the branch for drag-and-drop reordering in the next step.
- **KNOWN CONCERN — URL length.** Filter params already produce long URLs; adding `cols=` compounds this. Investigation into the best mitigation is underway.
- Row selection persists across column visibility changes. Filter changes clear it — **Reset All** in the filter panel clears filters only; **Reset** in Manage Columns clears column choices only.
- The state model uses a **hybrid inclusion/exclusion** approach. None mode (explicit row picks) sends only the checked IDs as `pwsids[]` — inclusion. All mode with some unchecks sends only the excluded IDs as `exclude_pwsids[]` — exclusion. The server always receives the smallest possible representation of the user's intent: a few explicit picks or a few exclusions from an otherwise-complete set.
- GeoJSON export is intentionally full-fidelity (all columns). `cols=` scoping applies to CSV only.

#### Test cases will be noted in a comment on the #117 ticket.

---

### Row Selection
- [ ] All rows are checked by default on page load
- [ ] Unchecking a row on page 1, navigating to page 2 and back — row 1 remains unchecked
- [ ] Select All restores all checks; Deselect All unchecks everything
- [ ] Export badge shows `All` when everything is checked; shows a count when some are unchecked; shows `0` when in None mode with nothing checked
- [ ] Export button is green by default, grays out on Deselect All, and restores green on Select All
- [ ] Export button `title` tooltip appears on hover when disabled
- [ ] Changing a filter clears all selections (all rows re-checked)
- [ ] Changing column visibility does NOT clear selections

### Export
- [ ] CSV export with all rows selected downloads all rows in current sort order
- [ ] CSV export with some rows unchecked excludes those rows
- [ ] CSV export in None mode with a few rows checked exports only those rows
- [ ] CSV column headers and data match the currently visible columns (hidden columns absent)
- [ ] Sort by a non-default column, export — confirm downloaded CSV rows are in sorted order
- [ ] GeoJSON export downloads with all columns regardless of visible column state
- [ ] No export submission fires when Export button is disabled (None mode, nothing selected)

### Manage Columns — Category Grouping
- [ ] Categories render with correct labels and collapse/expand chevrons
- [ ] Categories are collapsed by default on panel open
- [ ] Clicking a category header expands/collapses its children
- [ ] Category toggle-all checkbox checks/unchecks all children; checking expands the group, unchecking collapses it
- [ ] Checking/unchecking individual children updates the category header (checked, indeterminate, or unchecked)
- [ ] Reopening the panel correctly reflects the current `cols=` URL state

### URL State Persistence
- [ ] Apply a filter — `cols=` and `sort=` remain in the URL and table does not lose column/sort choices
- [ ] Navigate to the map view and back — column and sort choices are preserved
- [ ] Column reset removes only `cols=` — filters and sort remain
- [ ] Filter reset removes only filter params — `cols=` and `sort=` remain

### General / Regression
- [ ] Table renders correctly with no `cols=` param (all columns visible)
- [ ] Table renders correctly with `cols=` set to only pinned columns (`cols=`)
- [ ] Pagination works correctly with row selections active
- [ ] Export button color is green in the default state (not blue/brand-primary)


lukechanning (dismissed):
Nice man, this is really clean! Looks awesome and seems to work well, I did my best to scan the changes and there's a lot of solid practices in here. Nothing jumps out, I say let's ship it 🤷 🚢

CSheesley:
> Nice man, this is really clean! Looks awesome and seems to work well, I did my best to scan the changes and there's a lot of solid practices in here. Nothing jumps out, I say let's ship it 🤷 🚢

Thanks for the review, much appreciated! I'll wait and see if there is any feedback from EPIC before merging in.

lukechanning (dismissed):

CSheesley:
Cleanup fixes will be handled here: #213

PR #204, opened 2026-06-03
[Fix] Reconfigure GeoJSON Export for Speed + Accuracy

lukechanning:
## Summary

Replaces the GeoJSON export path with a PostgreSQL-driven streaming implementation. The old approach loaded all matching records into Ruby, serialised properties in application code, and gzip-compressed the result before sending. The new approach pushes all serialisation into the database via `ST_AsGeoJSON` and `json_build_object`, streams the FeatureCollection to the client in cursor-keyed batches of 500, and eliminates the gzip layer entirely.

> In my experience playing around with it, this is FAST and can even export the entire data set in less than a minute

## Changes

- **Streaming GeoJSON export** — `to_geojson_stream` returns an Enumerator that yields JSON chunks directly to Rack, keeping memory usage flat for any result size. Records are fetched in keyset-paginated batches (`WHERE pwsid > $last`) rather than loaded all at once.
- **Database-side serialisation** — geometry encoding (`ST_AsGeoJSON`), property building (`jsonb_build_object`), and all JOIN logic run in PostgreSQL. Ruby no longer touches individual row values for GeoJSON output. `GEOJSON_PROPERTY_COLUMNS` constant centralises the column-to-property mapping.
- **Removed gzip compression** — `Content-Encoding: gzip` and the `ActiveSupport::Gzip` call are gone. The response streams uncompressed JSON; browsers and HTTP clients handle transfer encoding via standard `Accept-Encoding` negotiation.
- **Deferred `with_details` for GeoJSON** — the controller no longer eagerly loads all associations for GeoJSON exports, since the SQL query handles joins directly. `with_details` is still applied for CSV exports.
- **SQL safety hardening** — `properties_sql` uses `connection.quote` / `connection.quote_column_name` for all interpolated values; `each_slice` capped at 49 pairs to stay under PostgreSQL's 100-argument limit; `fetch_feature_batch` annotated with a safety comment; Brakeman false positive suppressed with a documented ignore entry.
- **Test coverage** — new batch-boundary regression test (stubs `BATCH_SIZE = 1` to exercise the pagination loop); fixed a fragile `.last` reference in an existing feature-count test; request specs updated to parse uncompressed JSON directly.

## Notes for reviewers

The Brakeman ignore entry (`config/brakeman.ignore`) suppresses a raw-SQL warning on `fetch_feature_batch`. The note in the ignore file explains why each interpolated fragment is safe — all values derive from frozen constants or ActiveRecord's parameterised query builder. Worth a quick read to confirm the reasoning holds.

The `properties_sql` result is memoised on the exporter instance, so the subquery SQL is built once per export. The `filtered_ids_sql` subquery is re-executed by PostgreSQL on every batch fetch, which is correct for correctness but worth noting if exports over very large filtered scopes show query-planning overhead in future.

## Related Ticket
https://github.com/Environmental-Policy-Innovation-Center/water-data-tool/issues/179

## Screenshot
Screenshot 2026-06-04 at 11 47 21 AM

### DIFF
--- a/app/exporters/public_water_system_exporter.rb
+++ b/app/exporters/public_water_system_exporter.rb
@@ -1,6 +1,45 @@
 require "csv"
 
 class PublicWaterSystemExporter
+  BATCH_SIZE = 500
+
+  # Exported GeoJSON property columns, grouped by SQL table alias.
+  # Each [alias, columns] pair expands to alias.col => alias.col entries.
+  # The one renamed property (total_bwn → bws.total_notices) is merged at the end.
+  GEOJSON_PROPERTY_COLUMNS = [
+    ["pws", %w[
+      pwsid pws_name detailed_facility_report stusps counties gw_sw_code
+      source_water_protection_code owner_type primacy_type is_wholesaler
+      is_school_or_daycare service_area_type area_sq_miles open_health_viol
+    ]],
+    ["vs", %w[
+      health_violations_5yr groundwater_rule_5yr surface_water_treatment_5yr
+      lead_and_copper_5yr radionuclides_5yr inorganic_chemicals_5yr
+      synthetic_organic_chemicals_5yr volatile_organic_chemicals_5yr
+      total_coliform_5yr stage_1_disinfectants_5yr stage_2_disinfectants_5yr
+      health_violations_10yr groundwater_rule_10yr surface_water_treatment_10yr
+      lead_and_copper_10yr radionuclides_10yr inorganic_chemicals_10yr
+      synthetic_organic_chemicals_10yr volatile_organic_chemicals_10yr
+      total_coliform_10yr stage_1_disinfectants_10yr stage_2_disinfectants_10yr
+      paperwork_violations_5yr paperwork_violations_10yr
+    ]],
+    ["d", %w[
+      total_population population_density poverty_rate unemployment_rate
+      median_household_income bachelors_degree_rate age_under_5_rate age_over_61_rate
+      poc_rate white_rate black_rate aian_rate napi_rate asian_rate hispanic_rate
+      other_race_rate mixed_race_rate most_common_rate_tier
+    ]],
+    ["td", %w[population_pct_change mhi_pct_change]],
+    ["ej", %w[cejst_disadvantaged_pct svi_overall_pctl cvi_overall_score]],
+    ["fs", %w[times_funded total_srf_assistance total_principal_forgiveness]],
+    ["wh", %w[
+      num_facilities permit_effluent_violations open_underground_storage_tanks
+      risk_management_plan_facilities impaired_streams_303d
+    ]]
+  ].each_with_object({}) { |(tbl, cols), h|
+    cols.each { |col| h[col] = "#{tbl}.#{col}" }
+  }.merge("total_bwn" => "bws.total_notices").freeze
+
   # Human-readable headers matching the legacy datatable_export.csv output.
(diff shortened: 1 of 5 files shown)


app/exporters/public_water_system_exporter.rb
```diff
@@ -1,6 +1,45 @@
 require "csv"
 
 class PublicWaterSystemExporter
+  BATCH_SIZE = 500
+
+  # Exported GeoJSON property columns, grouped by SQL table alias.
+  # Each [alias, columns] pair expands to alias.col => alias.col entries.
+  # The one renamed property (total_bwn → bws.total_notices) is merged at the end.
```
CSheesley, line 8: Q: I am not super familiar with the exporting flow - but do we know why we need to rename any property, and/or why this is the only one?

app/exporters/public_water_system_exporter.rb
```diff
@@ -63,14 +102,30 @@ def to_csv
...
+          y << row["feature"]
+          last_pwsid = row["pwsid"]
+        end
+        break if rows.ntuples < BATCH_SIZE
       end
-      {type: "Feature", geometry: geometry, properties: geojson_properties(pws)}
+      y << "]}"
     end
-    {type: "FeatureCollection", features: features}
   end
```
CSheesley, line 129: From what I can gather, `y` seems to be the conventional variable name for these 'yielder' objects that we iterate over - but  what would you think about renaming that variable to something more explicit like `stream` to match the method name and the 'output' object that we are building up?

app/exporters/public_water_system_exporter.rb
```diff
@@ -115,82 +170,55 @@ def csv_row(pws)
...
-    td = pws.trend_datum
-    ej = pws.environmental_justice
-    fs = pws.funding_summary
-    wh = pws.watershed_hazard
+  def properties_sql
+    # PostgreSQL's json_build_object caps at 100 arguments (50 key-value pairs).
+    # Slice at 49 pairs to stay safely under that limit; merge chunks via the
+    # JSONB || operator then cast back to json.
+    # All keys and column references come from frozen constants — never extend
+    # this with runtime/user-supplied values.
```
CSheesley, line 178: This is a good call out - do we know if using 50 would be fragile or break?

CSheesley (dismissed):
I think this looks great - thanks for taking this on!

Just to check, does the CSV export still work - _I am assuming so based on the areas you touched being focused on geojson work._.

Based on our URL management chat, and the idea to convert these calls to post requests - I think this will work fine _(just using ids to help scope things further)_ - just wanted to see if you can think of any red flags with that pending change coming up.

Nice work!!

app/exporters/public_water_system_exporter.rb
lukechanning: Totes! So the `total_bwn` exists because DB surfaces `bws.total_notices ` so the export key has to match what FE expects. I think this is the only one that does this? I might be way off base here, but it seems to work...

app/exporters/public_water_system_exporter.rb
lukechanning: Super solid, fix incoming!

app/exporters/public_water_system_exporter.rb
lukechanning: I think it's a risk 🤷 To play it safe, might as well cap it below, but also... it's probably fine 😆 Dealers choice I'd say!

CSheesley (approved):

PR #222, opened 2026-06-17
[Chore] Fix a Docker Zip Dep + Manual Cartographic Boundary Workflow

lukechanning:
## Summary

  Add a manual GitHub Actions workflow to refresh cartographic boundary reference data in ECS, and bake `unzip` into the runtime image so the boundary loader can actually run in deployed environments.

  ## Changes

  - Add a `workflow_dispatch` action for refreshing cartographic boundaries in staging or production without using ECS Exec.
  - Support an optional temporary scale-down of the web service when ECS capacity cannot run the one-off refresh task alongside the app.
  - Run a follow-up verification task that checks `cartographic_states`, `cartographic_counties`, and `cartographic_places` all contain rows, and publish the task details plus row-count output in the workflow
  summary.
  - Improve operational failure reporting by surfacing ECS task startup failures and container exit details directly in the workflow logs.
  - Install `unzip` in the application image so `CartographicBoundaries.load` can extract Census shapefiles in deployed containers.

  ## Notes for reviewers

  - This is intentionally narrower than a full ETL run: it targets only the cartographic reference tables needed for boundary rendering and click-a-state behavior.
  - The workflow assumes the deployed ECS task definition already has the application code needed to run `CartographicBoundaries.load`; staging should be redeployed from `main` before using the new action there.
  - I only ran local static verification for the workflow (`actionlint` and YAML parsing). I did not run the GitHub Action or AWS/ECS refresh path from this environment.

### DIFF
--- a/.github/workflows/refresh-cartographic-boundaries.yml
+++ b/.github/workflows/refresh-cartographic-boundaries.yml
@@ -1,0 +1,289 @@
+name: Refresh Cartographic Boundaries
+
+on:
+  workflow_dispatch:
+    inputs:
+      target_environment:
+        description: "Environment to refresh"
+        required: true
+        type: choice
+        options:
+          - staging
+          - production
+      scale_service_down:
+        description: "Temporarily scale the web service to zero while the one-off task runs"
+        required: true
+        type: boolean
+        default: false
+      confirm:
+        description: 'Type "refresh-<env>-boundaries"; append "-with-downtime" when scale_service_down is true'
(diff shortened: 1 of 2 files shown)


CSheesley (approved):
Looks good to me - I'll go ahead and merge so that we can test.

As far as other things ETL related, I noticed that we still have the hardcoded urls for Layers, and the `loaded?` check in **CartographicBoundaries**. Do we/should we remove those?

PR #223, opened 2026-06-17
[FEAT] Add Drag and Drop to Manage Columns

### LINKED ISSUE #180: [FEAT]: Columns Drag and Drop - add drag and drop to data table

CSheesley:
## Summary

Wires SortableJS drag-and-drop in the Manage Columns panel and fixes a gap in state persistence: previously, "Show Columns" serialized only visible column keys, so hidden column positions were lost on full page reload and panel order diverged from the table when both were customized. Column order and visibility are now persisted together in the existing `encoded=` URL blob.

## Changes

- **Full panel state in the blob** — the `cols` field now holds the complete panel sequence using a `-key` prefix for hidden columns (e.g. `counties,-pwsid,stusps`). Plain key = visible in the table; `-key` = hidden but position preserved. Omitted entirely when all columns are at YAML defaults.

- **`ColumnState` value object** — `ColumnRegistry.parse_column_state` parses the `cols` string once into two derived lists: `panel_col_keys` (full sequence with `-` prefix, drives panel order and checkbox state) and `visible_col_keys` (clean symbols, drives table headers and exports). Each consumer gets exactly what it needs from a single parse.

- **Drag handle affordance** — column rows and category header rows now wrap the drag icon in a focusable `<button>` with an `aria-label`, enabling keyboard and assistive-technology access to reordering.

- **Deferred-apply pattern** — dragging or unchecking updates the DOM only (draft state). "Show Columns" commits the change by writing the blob and reloading the table frame. Clicking outside discards the draft. "Reset" restores YAML default order, checks all columns, and removes `cols` from the blob.

- **Server-rendered panel order** — panel list order on open reflects the last full page load (server-rendered from the blob). The JS controller syncs checkbox state from the URL on open but does not rehydrate DOM order — idiomatic Hotwire.

- **Backward compat** — `parse_column_state` recognizes the legacy visible-only format (no `-` prefix) and synthesizes the hidden tail from the remaining selectable columns, so old bookmarked URLs continue to work.

- **Category drag constraints** — columns can be reordered within their category; cross-category drag is blocked (no Sortable `group` option). Category blocks themselves can be reordered via handles on the category header rows.

## Notes for reviewers

The `-` prefix convention is the central design decision — it lets a single comma-separated string carry both order and visibility without a separate encoding layer. The full rationale and annotated blob example are in `docs/decisions/URL_MANAGEMENT.md`.

The deferred-apply pattern (click outside = discard) is intentional. The panel is a staging area; nothing hits the URL or the table until "Show Columns" is clicked.


lukechanning (dismissed):
Nice! Looks some serious work, nice and punchy changes. I dig it! 🚀

PR #228, opened 2026-06-20
[FEAT] Misc Map Work - smoothing, zooming, urls

### LINKED ISSUE #119: Update the behavior of loading of the state shape files so they load as one consistent shape
Currently when the tool loads the shape file for each state for the first time, they load in pieces which is confusing because of the shapes of the boundaries of public water systems. The state shape files should load as a single shape.

Image

Definition of done:
- [ ] State shape files should consistently load as a single shape

CSheesley:
# Summary

Map interaction fixes for state hover rendering, state zoom-to-fit behavior, and a report close button fix. Changes are isolated to `map_controller.js`, the shared URL sync module, and the report layer. 

## Tickets
- #119 
- _Several un-ticketed fixes/changes_
  
### Map — State Hover: Smooth Fill Rendering

Previously, hovering a state caused a `setFilter` call that re-evaluated all loaded vector tile data tile-by-tile, producing a "patchy" fill that appeared in staggered chunks.

Replaced with `setFeatureState`, which writes a `hover: true` flag into each feature's GPU-side state store. Because all tile fragments of the same state share a stable FIPS geoid key (via promoteId), they all respond in the same render frame — no patches. Added a `100ms` debounce on `mouseleave` to prevent flicker when the cursor crosses tile-boundary gaps between adjacent states.

### Map — State Selection & Zoom-to-Fit

- Clicking a state now runs `querySourceFeatures`, collects all polygon fragments for that state, builds a true `LngLatBounds`, and calls `fitBounds` — centering the state in the viewport rather than centering on the raw click coordinate.
- Sidebar width is factored into fit padding so the state isn't obscured behind an open panel.
- `maxZoom` is capped just below the systems-entry zoom level, preventing small states (DE, RI, MA) from zooming past the point where the PWS click layer disappears.
- Hawaii uses an explicit bounds override covering the 8 main islands instead of the computed bounds, which would include the Northwestern Hawaiian Islands chain far to the northwest.
- Resetting to national view from a zoomed-in state eases out gracefully instead of snapping.
- Extracted shared `url_sync.js` module so state FIPS and filter params write to the URL through a single path.

## Report — Close Button & From-Map Detection
  
- `ReportsController` now checks whether the referer is a same-host page and sets `@from_map`.
- When `@from_map` is true, the report layout renders a close button `(history.back())` instead of thestatic "back to map" link — correct for users arriving from within the app vs. a direct URL.
- Removed a dead `back()` action from `report_controller.js` that was never reachable (the report standalone layout loads no JavaScript).

## URL Sync — Shared Module (refactor)

- Extracted `syncToUrl` from a private method in `filter_controller.js` into a shared `url_sync.js` module. 
  - Behavior is identical — writes the current `FilterState` (filters + columns) and existing sort params into `?encoded=` — but now callable from `map_controller.js` as well, so state selection correctly persists to the URL through the same path as filter changes.


lukechanning (dismissed):
This is SO sick my friend — I've got some adjustments to the map work I was doing that will go really well with this. 

Let's definitely try to merge this one in when possible, LGTM 🚀

CSheesley:
> This is SO sick my friend — I've got some adjustments to the map work I was doing that will go really well with this.
> 
> Let's definitely try to merge this one in when possible, LGTM 🚀

Thanks! Sorry, I forgot that you were working on map stuff too... I was trying to knock out a lot of random tickets late last week and ended up spending a little time in map land.

My work did change some behaviors from previous AC's, so I will see what that feedback is like before adjusting this and merging things in.

PR #234, opened 2026-06-22
[Fix][Refactor] Map Flow Changes from Feedback

### LINKED ISSUE #236: Enhancement: Modify the behavior for hovering over a state
### Team / Area

Frontend

### Requested By

Kameron

### Overview

Currently when you hover over a state:
1. it turns green 
2. you get a hint to "select a state to learn more"

I'd recommend updating the behavior to:
1. turn green
2. similar to a system hover, increase the outline of the state to be thicker (from 1 pt to 2 pts)
3. change the cursor to be a pointer

I think these are enough to indicate that the state is clickable vs. the repetitive hint with the text.

### Context

It's an enhancement

### Success Criteria

_No response_

### Dependencies / Blockers

_No response_

### Links / References

_No response_

### Notes

LMK if you have any questions.

lukechanning:
## Merge Note

We should really merge this awesome PR by @CSheesley before we do this one. I'll ensure the changes play nicely together, then we can merge it in: 

https://github.com/Environmental-Policy-Innovation-Center/water-data-tool/pull/228

## Summary

Fixes several UX gaps in the map flow identified in feedback review. The main themes are: service area polygons now appear and are clickable at national zoom levels (not just after a state is selected), state hover/click interactions are correctly guarded once systems mode is active, and the cartographic boundaries loader is aligned with the standard ETL pattern (sourcing files from `ETL_SOURCE_URL` and recording a `DataImport`).

## Changes

- **National-scope service area visibility** — PWS polygons now render at all zoom levels, including before a state is selected. Clicking a service area at national scope selects its state and zooms to state-entry level rather than doing nothing.

- **PWS click deduplication and fallback** — PWS layer clicks are now deduplicated with a `__wdtPwsClickHandled` flag to prevent double-handling when the same click fires on both the `pws` layer and the generic map `click` event. A `queryRenderedFeatures` fallback catches clicks on rendered polygons whose event was not delivered directly by the layer handler.

- **State hover and click guards** — The selected state no longer shows a hover highlight or accepts a click while in systems mode. Hovering the selected state in systems mode clears any stale hover fill rather than applying a new one.

- **State hover prompt padding** — Added `px-4 py-3` to the "Select a state" popup so the text has breathing room.

- **Low-zoom tile generation includes PWS layer** — `TileGenerator.layers_for_zoom` now returns `pws` alongside `states` at national zoom levels. A dedicated `layer_simplification_tolerance` method gives PWS polygons a finer geometry tolerance at the national/state boundary zoom so they remain legible.

- **Cartographic boundaries sourced from ETL_SOURCE_URL** — Boundary zip files are now fetched from `ETL_SOURCE_URL/cartographic-boundaries/` using the existing `Etl::HttpFetcher` mixin, matching how all other ETL assets are fetched. Shapefile names no longer carry the Census year prefix. Successful loads record a `DataImport` row.

- **Lazy PostgreSQL array OID** — Replaced the `PWSID_ARRAY_TYPE` class-level constant in `PostImportSteps` with a method, so the PostgreSQL OID is resolved at call time rather than at file load, preventing boot-time errors in environments where the adapter isn't ready.

## Notes for reviewers

The PWS click deduplication (`__wdtPwsClickHandled`) is a property stamped directly onto the Mapbox event object. This is intentional — Mapbox GL fires both the layer-specific `click:pws` and the global `click` event for the same user click, and there's no API-supported way to stop propagation between them. The flag prevents double-handling without suppressing other handlers.

The `layer_simplification_tolerance` change means PWS tiles at z=4 are generated with `0.01` tolerance (same as z=5) rather than the coarser `0.05` used for states at that zoom. Tile sizes at z=4 will be slightly larger as a result.


app/services/cartographic_boundaries.rb
```diff
@@ -51,17 +56,22 @@ def load
...
     LAYERS.each { |layer| load_layer(layer, tmp_dir, conn, pg_conn_string, pg_password) }
+    record_import
   end
 
   private
 
+  def record_import
+    DataImport.create!(file_url: IMPORT_FILE_URL, imported_at: Time.current)
+  end
+
```
CSheesley, line 68: This looks good - is there any reason to make this a global helper used in all Importers, and somehow ensure that it always gets called after a successful import of any file type?

CSheesley (commented):
This looks good to me, thanks for wiring stuff up!

Two things

1. Just left a question about a potentially having the DataImport recording method being hoisted up, and called for ALL successful imports.

2. I am not sur where the change lives, but it looks like the zoom to state behavior is a little different on this branch versus what is on staging _(possibly by design)_.

Maybe check out staging and click on a state like NJ or CN to see the diff? On staging, I think our state level zooms are one set best fit view port - on this pr branch it looks like a state view sometimes vary (click NJ, click a neighboring state, click back to NJ - are the two NJ zoom views different)?

lukechanning:
> 1. Just left a question about a potentially having the DataImport recording method being hoisted up, and called for ALL successful imports.

Mm, good callout. So at the moment, we mostly get this through `Etl::FileImporter#call`, we record success after `import!. That being said, I do think there's a few locales where we do custom manual loads that skip this? Feel free to throw any specifics at me and I'll fix it up!

> 2. I am not sur where the change lives, but it looks like the zoom to state behavior is a little different on this branch versus what is on staging _(possibly by design)_.

Totally valid, fix is incoming!

re: @CSheesley

app/services/etl/post_import_steps.rb
```diff
@@ -20,13 +26,16 @@ def call(imported_files: nil, import_results: nil)
       if geometry_pwsids.any?
         fix_invalid_geometries(pwsids: geometry_pwsids)
         generate_centroids(pwsids: geometry_pwsids)
-        CartographicBoundaries.load
+        CartographicBoundaries.load unless CartographicBoundaries.loaded?
```
CSheesley: Just flagging this `.loaded?` check - do we want/need to add this check back?

PR #240, opened 2026-06-30
[Refactor] Address ETL Pipeline Feedback for Staging & Prod

lukechanning:
## Relevant Ticket
https://github.com/Environmental-Policy-Innovation-Center/water-data-tool/issues/199

## Summary

Moves environment-specific S3 behavior out of hard-coded staging URLs so ECS staging, production, and preview can all run with `RAILS_ENV=production` while still using the correct deployment identity. Adds app-level config for public downloads, methodology links, ETL source folders, and the recurring import gate.

## Changes

- Centralizes public asset URL configuration behind `AppConfig`, with `PUBLIC_DOWNLOADS_BASE_URL` and `METHODOLOGY_PDF_URL` taking precedence over environment-derived defaults.
- Updates the Downloads and Documentation links to resolve from runtime configuration instead of fixed `public-data-downloads/staged` paths.
- Aligns ETL source configuration with the AC for separate S3 folders: staging reads from `national-dw-tool/staging`, production reads from `national-dw-tool/prod`, and tests now assert both URL shapes.
- Uses `APP_ENV` as the app-level deployment identity so staging, production, and preview can behave differently even when ECS runs them all with `RAILS_ENV=production`.
- Gates the recurring SolidQueue ETL schedule with `ETL_SCHEDULE_ENABLED=true`, so only the intended ECS service enqueues `EtlImportJob` while staging/preview can stay manual by default.
- Documents the required deployment variables and data-population strategy for staging, production, and preview, including the expected `PUBLIC_DOWNLOADS_BASE_URL`, `ETL_SOURCE_URL`, and `ETL_SCHEDULE_ENABLED` values.
- Adds coverage for the AC-critical paths: config fallback behavior, env-backed public links, rendered home-page download/documentation URLs, recurring schedule inclusion/exclusion, and staging/prod ETL source URL construction.

## Notes for reviewers

The main review focus is whether the deployment env var contract matches the desired operational setup: `APP_ENV` identifies staging/prod/preview behavior, `ETL_SOURCE_URL` controls import inputs, `PUBLIC_DOWNLOADS_BASE_URL` controls rendered ZIP links, and `ETL_SCHEDULE_ENABLED` is the only switch that enables recurring imports.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,8 +9,20 @@ MAPBOX_STYLE_URL=mapbox://styles/your-account/your-style-id
 
 # ETL pipeline source URL — base HTTPS URL to the S3 folder containing the source data files.
 # Required for both local seeding (bin/rails db:seed) and the full ETL pipeline (bin/rails etl:import).
-# A single S3 bucket serves all environments (development, staging, production)
-ETL_SOURCE_URL=https://tech-team-data.s3.us-east-1.amazonaws.com/national-dw-tool/test-staged
+# Staging ECS should use .../staging; production ECS should use .../prod.
+ETL_SOURCE_URL=https://tech-team-data.s3.us-east-1.amazonaws.com/national-dw-tool/staging
+
+# Runtime environment identity. ECS services all run Rails in production mode,
+# so APP_ENV distinguishes staging, production, and preview behavior.
+APP_ENV=staging
+
+# Public download/documentation URLs rendered by the app.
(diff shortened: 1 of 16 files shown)


CSheesley (approved):
Awesome work, thanks for tackling this! I was also working on drying up the download and methodology URLs, but like this approach better!

PR #241, opened 2026-06-30
[REFACTOR] Simplify Configuration - four-file config model

CSheesley:
# Summary

**TLDR**:
_Simplifies how we add/remove/edit data fields and filters in the app, with a single source of truth for 'data fields', and a total of **Four** configuration files that need to be updated when changes are made._

This branch replaces our old configuration system — two large, overlapping YAML files, plus a per-source-file Ruby importer for every dataset — with a **four-file config model** backed by a single field **manifest**. Instead of scattering a field's identity across `columns.yml`, `filters.yml`, hand-written importer classes, and inline ERB, each data field is now declared **once** in `config/fields.yml`, and the rest of the system (ETL ingest, table columns, filter menus, histograms, exports) is driven from that declaration.

The result: adding, editing, or removing a data field is now a config change, not a code change, in the common case. We deleted ~4,300 lines and added ~4,050, but the real win is that the remaining config has **one source of truth per concern** and clear, one-directional relationships between files.

## What this got us *away from*

- **`config/columns.yml` (914 lines) and `config/filters.yml` (311 lines)** — two big files that each re-declared overlapping facts about the same fields (labels, sources, coercions), so a single field change meant edits in multiple places that could silently drift.
- **`FilterRegistry` (173 lines) and a wall of custom ETL importers** — `cejst.rb`, `cvi.rb`, `ejscreen.rb`, `svi.rb`, `epa_sabs_xwalk.rb`, `national_bwn_highlevel_summary.rb`, `pwsid_funded_highlevel_summary.rb`, `xwalk_pct_change_10yr.rb`, etc.
  - Most were near-identical flat 1:1 CSV→table maps that existed only to restate what the config already knew.
- **~620 lines of hand-maintained filter-menu ERB** — every menu/filter control was spelled out by hand in `_filter_menus.html.erb` and kept in sync with a ~636-line imperative `filter_controller.js`.

## What this got us *to*

**Four** config files, each owning exactly one concern:

1. **`config/fields.yml` — the field manifest** (read by `FieldRegistry`). Owns what each data field *is*: its identity and up to four capability blocks — `source` (how it's ingested from a source file), `display` (how it renders as a table column), `filter` (how it's filtered), `histogram` (how it charts). A field earns its place by having at least one block; omitting a block turns that capability off. This is *not* the full DB schema — a field appears only when we choose to ingest and/or surface it.

2. **`config/filter_layout.yml` — filter menu arrangement** (read by `FilterLayout`). Owns *where* a filter appears: which menu/category it sits in, its nesting and order, and the menu/category copy. It references fields by key; it never restates what a field *is*. Also encodes the faceted **AND/OR** logic (OR within a category, AND across categories).

3. **`config/table_layout.yml` — table column arrangement** (read by `TableLayout`). The table mirror of `filter_layout.yml`. Owns which columns show, their order, and their column-picker grouping. **One-directional:** the layout points at the manifest, never the reverse — a manifest field not listed here is simply hidden, and a layout key with no `display:` block is skipped (caught by spec, never a 500).

4. **`config/tooltips.yml` — user-facing copy.** Centralizes tooltip/help text so wording changes don't touch structural config.

## Changes

1. **New field manifest + registries.** Added `config/fields.yml`, `FieldRegistry`, and `FieldDefinition` (`app/fields/`). The manifest is now the single source of truth for a field's identity, sort behavior, permitted params, ETL column mapping, and export columns.

2. **ETL is manifest-driven.** Replaced the per-file importer classes with a single `Etl::Importers::Generic` that maps a source file straight to one model using `FieldRegistry.etl_mapping`, casting each value via `TypeCaster`. Genuinely custom importers (those that *derive* values or write to multiple models) remain, and an `importer_coverage_spec` enforces the generic-vs-custom split so nothing falls through the cracks.

3. **Filter menus are generated, not hand-written.** `_filter_menus.html.erb` shrank from ~620 lines to a small set of partials (`_filter_category`, `_filter_control`, `_filter_range`, `_filter_multiselect`, `_filter_radio`, `_filter_bool`, `_filter_rate_tier`, etc.) rendered from `fields.yml × filter_layout.yml`. `filter_controller.js` dropped from ~636 lines to a DOM-driven controller reading a `data-filter-*` contract.

4. **Server-rendered filter state.** Filter menu state now hydrates from the URL on the server rather than being reconstructed in JS, so the menus reflect the active filters on first paint.

5. **`Filterable` became a layout-driven combiner.** It now reads categories from `filter_layout.yml` and applies **OR within a category, AND across categories**, replacing the old `FilterRegistry`-coupled logic (233 → cleaner, layout-driven).

6. **Table columns and exports derive from the manifest.** `TableLayout` + `ColumnRegistry` build columns from `table_layout.yml × fields.yml`; the exporter derives its SQL from the manifest (dropped hand-written `value_sql` + table aliases). Disambiguated the derived read path as `read_from` to avoid colliding with the ETL `source:` block.

7. **Removed dead weight.** Deleted `config/columns.yml`, `config/filters.yml`, `FilterRegistry`, the redundant importer classes and their specs, the Place-as-a-filter feature (`places_controller`, `place_autocomplete_controller.js`, related route/specs), and pruned inert config keys (`version:`, `backend_only:`, orphaned `filter:`/`histogram:` blocks on non-surfaced fields).

8. **Map + stats polish (folded in from QA).** Unified state/territory zoom-to-fit so a state click and a region button frame identically via `#fitToState`; AK/GU/MP fly to fixed cameras where a bounding box can't frame them; added a `[busy]` dimming state on the stats bar for slow deployment round-trips.

9. **Docs.** Added `docs/CONFIG_AUDIT.md` (the full four-file design + phased migration record), updated `FILTERING.md`, `ETL.md`, and `URL_MANAGEMENT.md`, and retired the obsolete `FILTER_SERVER_RENDER.md`.

## Testing

Full suite green (~1,075 specs), including new parity specs for `FieldRegistry`, `FieldDefinition`, `FilterLayout`, `TableLayout`, and `importer_coverage`. `bin/ci` (standardrb, erb_lint, brakeman) passes.

## Follow-up (separate batch)

A set of **how-to docs for handling data fields** is in progress and will land in a later commit/branch, not this PR: `ADD_NEW_DATA_FIELD.md`, `EDIT_EXISTING_DATA_FIELD.md`, and `REMOVE_EXISTING_DATA_FIELD.md`, plus supporting guides (`GETTING_STARTED.md`, `COMMON_RAILS_COMMANDS.md`, `FIX_BIN_CI_CHECK_FAILURES.md`). These walk a contributor through the new manifest + layout workflow end-to-end so the four-file model is approachable without reading the source.


lukechanning (dismissed):
Dang dude, calm down! 😆 This is awesome, so much good work here. Obviously approved, cheers 🚀

PR #243, opened 2026-07-07
[FIX] Point downloads at the single existing S3 location

CSheesley:
# Summary

The merge of the ETL/AppConfig work pointed download links at a per-environment S3 layout (`public-data-downloads/{staging,prod}/national-dw-tool.zip`) that does not exist in the bucket, so **every download returned 403 AccessDenied** on local and staging (and would break prod on the next deploy). Downloads actually live in a single, environment-agnostic location _(confirmed by EPIC)_. This restores links to that location.          

## Changes

- `AppConfig.public_downloads_base_url` defaults to `.../public-data-downloads/staged`
- National download link uses the real object name `national-dw-tool-staged.zip`
- Removed the dead `s3_environment_folder` per-env switch and its only caller,`AppConfig.app_env`
- Removed the now unused `APP_ENV` environment variable from `.env.example`
- `.env.example`: corrected `PUBLIC_DOWNLOADS_BASE_URL` and its comment
- Updated `app_config_spec` to assert the single `staged` default

State downloads and the methodology PDF were already correct and are unchanged.
No ETL code is touched — `ETL_SOURCE_URL` resolution is unchanged _(needs to be set for each ENV)_.

## Testing
- Manual: Downloads panel — National, a state, and the methodology PDF all download (200)
- `bin/ci` and updated specs pass

## Note

This reverses the per-environment download-folder design — those folders don't exist in the bucket. If a bucket migration to per-env download folders was intended, flag it; otherwise single-location is the intent.

Follow up is needed to ensure we are setting `PUBLIC_DOWNLOADS_BASE_URL`, `METHODOLOGY_PDF_URL`, and `ETL_SOURCE_URL` correctly in our three environments.

Downloads - Environment agnostic - point at same bucket
ETL Source - Environment dependent
  - Prod -> prod
  - Staging & Preview -> staging

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -12,13 +12,9 @@ MAPBOX_STYLE_URL=mapbox://styles/your-account/your-style-id
 # Staging ECS should use .../staging; production ECS should use .../prod.
 ETL_SOURCE_URL=https://tech-team-data.s3.us-east-1.amazonaws.com/national-dw-tool/staging
 
-# Runtime environment identity. ECS services all run Rails in production mode,
-# so APP_ENV distinguishes staging, production, and preview behavior.
-APP_ENV=staging
-
 # Public download/documentation URLs rendered by the app.
-# Staging ECS should use .../public-data-downloads/staging; production ECS should use .../public-data-downloads/prod.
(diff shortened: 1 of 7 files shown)


lukechanning (approved):
👍 🚀

PR #254, opened 2026-07-15
Added AWIA Data

EmmaLiTsai:
Added AWIA data!

### DIFF
--- a/app/models/certification_summary.rb
+++ b/app/models/certification_summary.rb
@@ -3,6 +3,7 @@
 # Table name: certification_summaries
 #
 #  id                :bigint           not null, primary key
+#  erp_certification :string
 #  pwsid             :string           not null
 #  rra_certification :string
 #  created_at        :datetime         not null
--- a/config/fields.yml
+++ b/config/fields.yml
@@ -744,3 +744,14 @@ fields:
       param: rra_certification
       checked_value: Certified
       label: "Risk and resilience assessment"
+
+  erp_certification:
+    model: certification_summary
+    source: {file: awia_certification, header: "erp_certification", cast: string}
+    display: {label: "Emergency response plan", sort: erp_certification, format: str, csv_label: "Emergency response plan"}
+    filter:
+      kind: bool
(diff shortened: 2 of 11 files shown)


CSheesley (approved):
Nice job 👍

EmmaLiTsai:
@CSheesley Thanks! Any idea why the new certification data are not showing up in the preview environment? I was thinking it would appear after the ETL pipeline ran overnight but just double-checking

EmmaLiTsai:
@KameronKerger - this fixes the histogram filter bug (thanks, Corey!) and also adds the AWIA datasets. A note: I still need to add the AWIA datasets to the datasets tab, the zipped data downloads, public methods, and data dictionaries. My hope is that we can push this to staging, and once that seems solid, promote to production so we can fix that histogram filter bug. 

I still need to add the AWIA dataset card to the datasets tab of the tool & will tag you in for a review! [update: it was easy & has been updated! ready for your review!]

CSheesley:
> @CSheesley Thanks! Any idea why the new certification data are not showing up in the preview environment? I was thinking it would appear after the ETL pipeline ran overnight but just double-checking

Hey @EmmaLiTsai - just now seeing this. I noticed the same thing when trying test my part of the AWIA add ticket. That ETL for the Preview ENV never successfully ran, even when I tried to manually run the workflow that kicks off an ETL run in the Preview ENV _(which would be used to manually update the preview db for quicker testing, without having to wait for a nightly run)_: https://github.com/Environmental-Policy-Innovation-Center/water-data-tool/actions/runs/29400530060

I am not as well versed on the infrastructure side of things, but I think this might be do to an 'instance size' or memory issue for the task that runs that job for the preview ENV. That said, maybe bumping the size of the instance that runs preview tasks could help?

All that said - It did look like the ETL ran successfully on Staging, and you should be able to see the AWIA data there (at least for piece that I added). If you merge your part in and wait a day, the data that your AWIA add relies on should populate as well.

KameronKerger:
🎉

